Issue #55, opened 2014-08-19
Support Extended Wildcards


rnc:
It would be useful to have extended support for wildcards in exclusions to also include artifactId i.e. so rather than
```
properties=pluginManagement=,
   dependencyExclusion.org.dummy.groupId:artifactId-common@*=,
   dependencyExclusion.org.dummy.groupId:artifactId-bus@*=,
   dependencyExclusion.org.dummy.groupId:artifactId-ioc@*=
```
we could have
```
properties=pluginManagement=,
   dependencyExclusion.org.dummy.groupId:*@*=
```
which would much simplify the command line and avoid aligning anything with that groupId in any module.

rnc:
Fixed by #114

PR #114, opened 2015-02-25
Issue 55

rnc:


rnc:
For for issue #55

jdcasey:
This looks pretty good, but I wonder if we're going to need to support partially wildcarded groupIds too, like:

  - `org.jboss.as.*:*@*`

rnc:
*Maybe* ... I've only had a request for wildcard on the artifact so far. If we did need that would it not be a matter of changing the implementation within WildCardMap i.e. its mostly encapsulated at the moment ? The only bit that isn't is https://github.com/release-engineering/pom-manipulation-ext/pull/114/files#diff-4bbf6a93e6938769dff5f0f6d9d0fe32R645

jdcasey:
yep, you're right, it's very well contained. I'm happy with this as a starting point.

Issue #122, opened 2015-04-09
REGRESSION: versions no longer zero-filled for matching


jdcasey:
given POM version: '7'
given metadata versions:
  * '7'
  * '7.0.0.build-3'

The new version should be '7.0.0.build-4', but it's '7.0.0.build-1' (it's not zero-filling the version when matching against the metadata versions). I've confirmed that this worked when we were using atlas for version parsing / sorting, but broke when we moved to the internal Version class for parsing/sorting.

PR #123, opened 2015-04-09
adding test for regression in #122

jdcasey:
Also adding a ModelSource implementation that uses the galley Transfer object directly.

### DIFF
--- a/src/test/java/org/commonjava/maven/ext/manip/impl/VersionTest.java
+++ b/src/test/java/org/commonjava/maven/ext/manip/impl/VersionTest.java
@@ -16,6 +16,7 @@ import static org.junit.Assert.assertThat;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.junit.Ignore;
 import org.junit.Test;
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/commonjava/maven/ext/manip/impl/VersionTest.java
```diff
@@ -265,7 +266,17 @@ public void testFindHighestMatchingBuildNumber()
         versionSet.add( "1.2-foo-4" );
         assertThat( version.findHighestMatchingBuildNumber( version, versionSet ), equalTo( 4 ) );
         versionSet.clear();
+    }
 
+    @Test
+    @Ignore( "PENDING: https://github.com/release-engineering/pom-manipulation-ext/issues/122" )
```
rnc: Should this be @Ignore as #122 has been closed now ?

jdcasey:
The @Ignore was cleared once the code was fixed...I didn't want to leave the repo in a non-building state.

PR #124, opened 2015-04-10
[issue122] Fix cases with a major only version

pgier:
If a simple major only version was previously converted to OSGi and then incremented
with a suffix, make sure we can match that in the existing set of versions in the repository.


jdcasey:
Fixes #122 (let's see if this automatically closes the issue)

Issue #121, opened 2015-04-08
Detect use of ${project.version} in dependencies


jdcasey:
This can be especially dangerous if the POM is a BOM (or a parent POM that someone might inherit from outside of its own project).

When either of these are used, the consuming project's version is used for ${project.version}, NOT the intended version from the original BOM/POM.

We should be able to detect this - on both ends, producing and consuming - and take one of three actions:

* warn (producer side...this is weak, I'm not a fan)
* fail the build (producer side)
* correct seamlessly (consumer and producer)

jdcasey:
if packaging != pom, we can probably ignore this issue...you can't inherit from a POM unless it's got that packaging, and I would hope that the same holds for BOMs (though I haven't verified it in the Maven codebase).

rnc:
Regarding warn - given the amount of output already it is likely warnings would be missed.

What about failing the build if we detect the consumption of an invalid [use of project.version] POM ? 

Surely if the producer is corrected during the build, the consumer would not have to be corrected?...it would be transparent?

jdcasey:
Agreed, the simplest fix would be to fix on the producer side and let the consumption just detect/fail. The only problem is it won't address the use of older artifacts/POMs.

Also, it might be worth examining if this still causes problems on new versions of Maven. Someone mentioned to me that they couldn't reproduce the problem from my description when using a recent version.

rnc:
I have tested with Maven 3.2.5. I can reproduce the problem when using inheritance but not when using import.

rnc:
Fixed by #125 and #125

PR #125, opened 2015-04-21
Issue 121: Fix up project.version in pom files

rnc:
This is for issue #121.

Currently I have configured it to automatically fix the build by default (and it is possible to turn this off).

### DIFF
--- a/src/main/java/org/commonjava/maven/ext/manip/impl/ProjectVersionEnforcingManipulator.java
+++ b/src/main/java/org/commonjava/maven/ext/manip/impl/ProjectVersionEnforcingManipulator.java
@@ -1,0 +1,140 @@
+package org.commonjava.maven.ext.manip.impl;
+
+import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Model;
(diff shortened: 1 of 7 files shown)


src/main/java/org/commonjava/maven/ext/manip/impl/ProjectVersionEnforcingManipulator.java
```diff
@@ -0,0 +1,140 @@
...
+        final Set<Project> changed = new HashSet<Project>();
+
+        for ( final Project project : projects )
+        {
+            final String ga = ga( project );
+            final Model model = project.getModel();
+
+            if ( model.getPackaging().equals( "pom" ) )
+            {
+                if ( state.isEnabled() != true )
```
jdcasey: Is this redundant since you have a short circuit above in the method?

rnc: Good point. I will fix this in a new pull request

PR #170, opened 2015-07-27
Update Javadoc. Minor API cleanup. Update to Galley 0.9.1

rnc:


core/pom.xml
```diff
@@ -97,6 +97,12 @@
       <artifactId>atlas-identities</artifactId>
     </dependency>
 
+    <!-- Can't build javadoc in 8 without this dependency -->
+    <dependency>
+      <groupId>org.commonjava.cdi.util</groupId>
+      <artifactId>weft</artifactId>
+    </dependency>
+
```
rnc: @jdcasey : Not sure about this : couldn't get it to build under JDK 8 without adding this.

Issue #177, opened 2015-07-30
Consider refactoring integration tests into their own module


rnc:
Ensure they can be run in either ext mode or in CLI mode.

jdcasey:
We could put the tests themselves (project directory content) in a separate module, then provide test "drivers" that reference these projects and run the builds...for the ext, it'd probably still use invoker, but the cli would probably use normal junit tests.

Is that what you have in mind?

rnc:
Yes pretty much -  a integration-test module, with perhaps invoker for ext tests and some other form of driver for the CLI - ideally all based on the same set of tests. Although we might end up needed exclusion/inclusion in case some tests are not applicable.

rnc:
I think the createSession in Cli needs to support setting a local repository (and plugin repository?) .... i.e. equivalent to the Maven -s command.

VaclavDedik:
So I am currently using Junit to execute the tests in the cli mode. I am reusing basically everything in the "it" directory (test.properties, pom.xml, verify.groovy), there is a problem though. When I execute the PME in cli mode, it doesn't do "mvn install" as maven-invoker-plugin does. So when the "verify.groovy" is executed, the test fails because some of the integration tests check for jar (and pom).

One solution I can think of is to remove these checks from the tests. In my opinion this wouldn't be such a big deal because I think the tests should test if PME works correctly, not if "mvn install" works correctly.

Another solution would be to run only some tests in the cli mode.

What do you think?

jdcasey:
I think I'd prefer to exclude those tests from the CLI suite's execution, IF the tests are really checking something related to the build itself (for instance, that maven is actually using the changed POMs). Otherwise, I'd say modifying the verify.groovy scripts would be fine.

If we do run into tests that have important verifications but *also* really need to verify output from a maven build, I'd suggest cloning them and creating a version that only verifies the maven install output (which would be excluded from CLI tests), and a version that only verifies the other stuff...basically decomposing the tests. Tests like that are a sort of bad code smell anyway, in that they wouldn't really be testing a single feature/behavior.

rnc:
From taking a quick look at the groovy scripts they mostly seem to be doing this kind of thing
```
def jar = new File(basedir, "target/${pom.artifactId.text()}-${pom.version.text()}.jar" )
System.out.println( "Checking for jar: ${jar.getAbsolutePath()}")
assert jar.exists()

[...]

PR #180, opened 2015-08-05
Issue #177 -- Move integration tests into integration-test module

VaclavDedik:
This commit moves the integration tests from ext module to integration-test module. I hope I've done it right, I replaced the maven-resources-plugin that copied the build extension into ${it.maven.home}/lib/ext wtih maven-dependency-plugin, not sure if this is the best solution but I couldn't think of any other.


integration-test/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+    <dependency>
+      <groupId>org.commonjava.maven.ext</groupId>
+      <artifactId>pom-manipulation-ext</artifactId>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
+      <id>run-its</id>
+
```
rnc: It might be simpler to have the profiling in the modules in the parent pom and keep this one simple.

rnc:
Don't we also need to modify the top level pom.xml to add the new integration-test module - so we can build everything ?

VaclavDedik:
Yes we do, and I did change the top level pom.xml.. it seems github can't show the entire diff because it is too big. You can see that the top level pom changed by looking at for example [this](https://github.com/VaclavDedik/pom-manipulation-ext/blob/master/pom.xml#L414) line in my fork.

integration-test/pom.xml
VaclavDedik: You are right, it is probably better, I'll move the profile.

rnc:
@VaclavDedik : Re top pom , apologies missed that.

rnc:
test this please

VaclavDedik:
I made the profile modification you suggested, I can see that it makes much more sense this way.

PR #185, opened 2015-08-14
Issue #177 - Fix integration tests in CLI mode

VaclavDedik:


jdcasey:
retest this please

jdcasey:
Is it possible you wrote some of these tests using a JDK > 1.6?

See: http://ci.commonjava.org:8080/job/pom-manipulation-ext/59/testReport/

VaclavDedik:
Yes, I am using JDK 1.7 on my laptop. I was careful not to use anything Java 7 specific, but I guess it doesn't matter, because the problem is with an external library that is (probably) compiled with JDK 1.7?

VaclavDedik:
No, it actually seems I must have switched to 1.8 (because otherwise maven would fail because of [enforcer rule] (https://github.com/release-engineering/redhat-releng-tools/blob/master/pom.xml)). Not currently at my business laptop so can't verify, but I must be running 1.8.

jdcasey:
Currently, we have users still on JDK 1.6, so we need to make sure this can build with jdk 1.6...otherwise, we're stuck using animal-sniffer to ensure nothing like this slips into the jar we distribute. I sort of doubt animal-sniffer would have picked up on a 1.7+ dependency.

Would it be possible for you to install JDK 1.6 on your machine too, and use it for developing PME? I have a couple of scripts that I use for this when I do command-line builds. I can simply run:

    $ export JDK=1.6

and it'll select the appropriate java binary / JAVA_HOME. If you want, I can pass them along. Also, Eclipse and Idea both support project-specific JDK versions.

VaclavDedik:
Yes, I can install JDK 1.6, no problem. I usually switch versions with update-alternatives so it's no problem for me.
But the maven build will fail if I switch to anything other than JDK 1.8, won't it? I am quite sure it failed when I tried to build PME with JDK 1.7 (complaining about java version) and therefore I switched to JDK 1.8. The parent pom of the top level pom of PME [here](https://github.com/release-engineering/redhat-releng-tools/blob/master/pom.xml) seems to have an enforcer rule for JDK 1.8.

jdcasey:
I just built my local copy using JDK 1.6, though I did have to use -DskipTests=true to avoid problems with jetty embedded in the tests. The enforcer rule should use ${javaVersion}, which in PME is set to 1.6.

VaclavDedik:
I see, you are right, in the PME pom it is set to 1.6. I wonder why my build failed when I used 1.7 and why it works when I use 1.8.

Anyway, I'll try to come up with a solution to this problem on Monday. Either I'll try to replace Jetty with an older version that is built in 1.6 or I'll have to replace rest-driver with something else.

VaclavDedik:
I finally managed to fix the problem. I tried to replace jetty dependencies of rest-driver with different version but that (not that surprisingly) didn't work. So I tried to use wiremock instead, but I didn't manage to make it work. So I tried mockserver and again, couldn't make it work. Lastly I tried betamax, but that didn't seem to be a good framework for this problem. So eventually I just decided to use Jetty directly (i.e. start it in embedded mode from java code with handler that appends 'redhat-1' suffix to all incoming requests). This actually seems even better than rest-driver.

So now it should be working, I set my java to oracle java 1.6 and all tests passed and the whole project was successfully built.

jdcasey:
retest this please

jdcasey:
@VaclavDedik sounds great. I'll wait for this CI build to finish, then merge.

Thanks!

VaclavDedik:
Weird. It downloaded jetty 9.1.1 even though I set the version to 8.1.. I'll look into it.

jdcasey:
Thanks. I was just looking at that on my localhost, trying to figure it out. But somehow, when I build from your master branch using JDK 1.6, things seem to freeze at:

Running org.commonjava.maven.ext.manip.DefaultCliIntegrationTest

I'm not sure what to make of that, given that it runs on the CI system.

VaclavDedik:
Hm, are you sure it freezes on it? It takes a lot of time to finish (about 6 minutes on my laptop).. and unfortunately even though I use parametrized tests, it doesn't show any progress so you only know which parametrized tests finished after all of them finish.

jdcasey:
hahaha, you're right. I just went back and looked at that shell again, and it succeeded :)

Ok, so I think the CI build itself might be broken in a way. I've seen in the past that sometimes changes in the POM via PR don't always get picked up. I'll merge this and we can see if the CI build is still broken. I suspect it's nothing on your end of things.

VaclavDedik:
I see, cool... fingers crossed :)

jdcasey:
Looks like we're good. Thanks again.

PR #190, opened 2015-08-27
Rewrite VersionTranslator, refactor mock server, reformat code

VaclavDedik:


VaclavDedik:
Slow Jenkins. On my laptop it takes the performance test about 59 miliseconds, on jenkins it takes 329 miliseconds... I increased it to 500 ms.

PR #217, opened 2015-09-08
disable addition of version-override properties, which uses VersionPropertyFormat

jdcasey:
...in anticipation of removal for this section of the code.

@rnc The build succeeds on my localhost after this change. Do you know if this section is really unused now, or if the tests are just a little light?

I'm happy to continue work on removal, but wanted to ping you about this first.


rnc:
@jdcasey  : the only test appears to be integration-test/src/it/property-injection-disable (testing that it is disabled it seems)

PR #231, opened 2015-10-07
Handle scenario where there is a single version.suffix override and i…

rnc:
…t has leading zeros


rnc:
@jdcasey : once this is merged can you do a new release please?

Issue #229, opened 2015-10-02
Missing documentation of feature "dependency manipulation - REST Endpoint"


torsten-liermann:
Hi,

(sorry, I can't speak english). I'm missing an example for  the feature "REST Endpoint". What is the property name for specifying the rest URL? 

Thanks.

rnc:
Good point. Its currently specified as `-DrestURL` in https://release-engineering.github.io/pom-manipulation-ext/guide/project-version-manip.html but we should probably cross-reference it in https://release-engineering.github.io/pom-manipulation-ext/guide/dep-manip.html as well.

Note that PME won't be able to use the endpoint unless it has been set up. The project that provides the endpoint has been cross-referenced (https://github.com/project-ncl/dependency-analysis / https://github.com/project-ncl/dependency-analysis/tree/master/reports-rest).

rnc:
@torsten-liermann : what are you trying to do with PME ? Can we help ?

torsten-liermann:
may be You can help me. Please contact me via torsten@liermann-it.de 
Thanks!

torsten-liermann:
Hi Nick, thanks for the useful project. Is pom-manipulation-ext open for extensions or must I fork for implementing own ideas? I would find it better if we could talk about ideas.
Thanks,
Torsten

jdcasey:
So, are you using snapshot versions in between releases? If so, I guess you'd need to tag the release and then re-version your project back to the next snapshot, similarly to how the release plugin does now?

rnc:
Hi Torsten
I think both myself and @jdcasey have sent you an email re your usage. We'd be happy to look at any ideas you have.

torsten-liermann:
@jdcasey Just as you suggest it, we do it for years. But taking care of dependencies between projects is error-prone. The idea of resrURL might be a more elegant solution.

rnc:
With the additional documentation I think this is done. Feel free to reopen if not :-)

torsten-liermann:
Thanks! Thanks! What is the idea behind PME also expect the description of the reactor artifacts?

PR #233, opened 2015-10-08
Add more details on restURL. Add repo-removal-ignorelocalhost

rnc:


rnc:
This should also resolve #229

Issue #236, opened 2015-10-19
Plugin injection / properties


rnc:
Ideally remote plugin injection should use the same method as the dependency manipulation when it encounters a property - follow that property and update its value rather than injecting the value directly.

jdcasey:
Merged associated PR. Thanks!

PR #238, opened 2015-10-19
Issue #236 Update Plugin properties as per dependencies

rnc:


core/src/main/java/org/commonjava/maven/ext/manip/impl/PluginManipulator.java
```diff
@@ -262,14 +295,45 @@ else if ( configPrecedence == Precedence.LOCAL )
...
-                if ( pluginMgmt || plugin.getVersion() != null )
+                if ( pluginMgmt || oldVersion != null )
                 {
-                    plugin.setVersion( override.getVersion() );
-                    logger.info( "Altered plugin version: " + groupIdArtifactId + "=" + override.getVersion());
-                }
+                    // Handle the situation where we are updating a dependency that has an existing property - in this
+                    // case we want to update the property instead.
+                    // TODO: Handle the scenario where the version might be ${....}${....}
+                    if ( oldVersion.startsWith( "${" ) )
```
jdcasey: It looks like this is something we could move to PropertyUtils, maybe something like:

    public static String findPropertyRef( String version);

I'm just guessing if it's here we probably have similar logic somewhere else, related to dependencies?

core/src/main/java/org/commonjava/maven/ext/manip/impl/PluginManipulator.java
```diff
@@ -63,6 +65,11 @@
     private Precedence configPrecedence = Precedence.REMOTE;
 
     /**
+     * Used to store mappings of old property to new version.
+     */
+    protected final HashMap<String, String> versionPropertyUpdateMap = new HashMap<String, String>();
```
jdcasey: It might be better to have this in PluginState, just to keep the state separated from these components. I don't think is an issue now, but if we ever turn PME into a standalone service that has something like REST endpoints, we won't want to keep state like this in the components. Pushing it into PluginState now avoids a gotcha if that ever comes about.

core/src/main/java/org/commonjava/maven/ext/manip/impl/PluginManipulator.java
rnc: You're right it is _similar_ ; the only difference is in the debug line e.g. https://github.com/release-engineering/pom-manipulation-ext/blob/master/core/src/main/java/org/commonjava/maven/ext/manip/impl/DependencyManipulator.java#L604

core/src/main/java/org/commonjava/maven/ext/manip/impl/PluginManipulator.java
rnc: For reference - DependencyManipulator also has two local hashmaps it uses to track changes as it progresses through.

PR #248, opened 2015-10-27
Add GroovyManipulator

rnc:
@jdcasey I think this can be reviewed now ... (still have the gh-pages docs to write after this). Have added some comments inline.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
@@ -106,6 +106,8 @@ public class GroovyManipulator
                 {
                     binding.setProperty( "basedir", project.getPom().getParentFile().toString() );
                     binding.setProperty( "name", project.getKey() );
+                    binding.setProperty( "project", project );
+                    binding.setProperty( "projects", projects );
 
                     logger.info ("Executing {} on {} with binding {} ", groovyScript, project, binding.getVariables());
 
--- a/core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
@@ -17,8 +17,10 @@ package org.commonjava.maven.ext.manip.impl;
 
 import groovy.lang.Binding;
 import groovy.lang.GroovyShell;
+import groovy.lang.MissingMethodException;
 import groovy.lang.Script;
 import org.codehaus.groovy.control.CompilationFailedException;
+import org.codehaus.groovy.control.CompilerConfiguration;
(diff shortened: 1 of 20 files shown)


core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+        return changed;
+    }
+
+
+    @Override
+    public int getExecutionIndex()
+    {
+        return 90;
+    }
```
rnc: Currently it runs after all other manipulations.

core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
```diff
@@ -0,0 +1,146 @@
...
+            GroovyShell shell = new GroovyShell( binding );
+            Script script = null;
+
+            for ( final Project project : projects )
+            {
+                if ( project.isExecutionRoot() )
+                {
+                    binding.setProperty( "basedir", project.getPom().getParentFile().toString() );
+                    binding.setProperty( "name", project.getKey() );
+
```
rnc: I could inject a lot more information ... at the moment just the execution root directory and name ... should we inject anything else? Note we only run it once - at the execution root.

pom.xml
```diff
@@ -302,9 +302,18 @@
 
       <dependency>
         <groupId>org.codehaus.groovy</groupId>
-        <artifactId>groovy-all</artifactId>
-        <version>2.0.1</version>
-        <scope>test</scope>
+        <artifactId>groovy</artifactId>
```
rnc: I switched to the smaller groovy jars so the extension jar isn't so large (shaved 6MB off that way)

integration-test/src/it/setup/depMgmt1/Sample.groovy
```diff
@@ -0,0 +1,56 @@
+
+import groovy.json.JsonOutput
+import groovy.json.JsonSlurper
+import groovy.util.logging.Slf4j
+
```
rnc: We're using this code to post-process the json files to remove the resolved lines.

core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
jdcasey: I'd probably bind the projects too, just to avoid any confusion in rewriting them once in groovy, then again in java...if that happened, the java rewrite would erase the changes from the groovy script, since the project in memory was read ahead of time.

jdcasey:
retest this please

jdcasey:
retest this please

jdcasey:
retest this please

jdcasey:
retest this please

Issue #257, opened 2015-11-10
Plugin config injection does not take into account properties in the local build


pgier:
The jboss-parent pom has plugin config for the maven compiler plugin which looks like this
```xml
  <properties>
    <maven.compiler.target>1.6</maven.compiler.target>
  </properties>
  ...
  <configuration>
    <target>${maven.compiler.target}</target>
  </configuration>
```
When this is injected into a build, the property is resolved to the value set in the jboss-parent pom and it ignores local system properties.
```xml
  <configuration>
    <target>1.6</target>
  </configuration>
```
This means that after plugin injection (compared to POM inheritance) I can no longer override this setting at run time.  When the value of ${maven.compiler.target} is resolved, it should take into account whether I set the local property "-Dmaven.compiler.target".

rnc:
Related : #91

rnc:
Should be fixed now

PR #261, opened 2015-11-12
Issue 258 Allow plugin executions to be injected.

rnc:


rnc:
@jdcasey Please don't merge this until #257 is merged (it layers on top of that). I might need to rebase...

rnc:
@pgier : would this satisfy your use-case for executions? Note that it will fail the build if there is a clash between the injected executions and existing executions.

jdcasey:
retest this please

Issue #237, opened 2015-10-19
set value of scm connection for projects checked out from git


psakar:
create pom manipulation for scm connection.

if manipulation is enabled and project was check out from git, than scm connection will be set to url read from git (url of remote origin stored in file) .git/config.

Information from scm connection can be used later eg. to write it to MANIFEST.MF

Added option to strip domain name from git host for case when security rules prohibit to publish internal domain names

rnc:
This is an interesting use case for when building a project that has been mirrored to secondary  storage (i.e. we are not building from the upstream SCM URL). 

Note that the build metadata plugin already stores a large amount of data about the project. It parses the POM and the current environment.

__However__ this information is stored within a `build.metadata` file which is collected and stored by the `project-sources-maven-plugin`. It is not stored within a jar file. 

So : 

1. Should we preserve the upstream information ?
2. Should we make the build.metadata plugin record this extra information ?
3. If we wish to try and embed information within a jar file as well - how we handle the differing use of maven-jar-plugin (not configured so using default manifest), a configured maven-jar-plugin, maven-bundle-plugin etc.

jdcasey:
Is there an actual use case for the actual information in the manifest now? Or are we building out some sort of capacity we think we might need in the future?

What's the problem we're trying to solve? If this is sensitive, we can take the discussion offline (email) for this point.

psakar:
it is part of effort to solve two problems
- fix the wrong information in MANIFEST.MF - it contains SCM-URL which is not the url from which the artefact was build)
- MANIFEST.MF contains SCM-URL containing internal hostname, publishing such information is forbidden by company security policy

rnc:
@psakar However an awful lot of builds do not have MANIFEST information. Especially those created by e.g. bundle, shade etc.... so just changing the pom file would not solve manifest (although it might solve the bundled pom inside the jar file).

rnc:
Discussed offline and its possible to write a plugin for the metadataplugin to extend its functionality. Prototype is https://github.com/psakar/buildmetadata-plugin-git-provider 

I'm closing this for now.

PR #262, opened 2015-11-12
Configure buildmetadata to write information to the META-INF director…

rnc:
…y as well.

Note: this needs a new release of buildmetadata plugin and the plugin version updating in `core/src/main/java/org/commonjava/maven/ext/manip/state/ProjectSourcesInjectingState.java`

This would also help with #237 as now the information is contained within the manifest. By using the metadata plugin SPI we could create an extension that would also store an abbreviated version of the internal SCM repository in the metadata file.


jdcasey:
retest this please

PR #263, opened 2015-11-12
Use port-finder for REST mock server, to avoid problems in CI.

jdcasey:
Also, wrap lower exceptions when throwing at a higher level (looking at you, DefaultVersionTranslator). That way we don't lose the stack trace


jdcasey:
retest this please

Issue #273, opened 2015-11-17
pluginManagement fails when run more than once on the same POM


pgier:
If pluginManagement injection is used on a POM, PME checks for conflicts between the injected config and the local POM config.  This is fine for the first execution of a build.  However, if PME is run a second time it conflicts with the previous injections even though the config is the same.
It would be good if the pluginManagement injection could compare the existing config, and if they are the same, then just skip that part of the injection and continue the build instead of failing.

rnc:
Are you referring to injected executions (which might contain configuration ) or injected configuration ?

pgier:
I consider the executions in pluginManagement part of the config, so I'm referring to both.  I guess it's a question of whether we want to be able to run PME twice against the same POM.  Currently this fails because PME thinks there is a conflict on the second run.  The main use case I can think of for running PME against a POM which already had it run is for rebuilds.  For example, if we check out sources which already had PME injections, and then run the build with the same CLI settings, then currently the build will fail.
I think it would be better if PME examines the conflicting config during execution and only fails if there is an actual difference.

rnc:
Fixed by #275

PR #275, opened 2015-11-18
Issue #273 Don't fail if attempt to inject twice

rnc:


rnc:
This is for issue #273

PR #278, opened 2015-11-20
Check for null version before trying to resolve property

pgier:
Checks the value of oldVersion parameter before trying to parse the string
to avoid NullPointerException

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/manip/util/PropertiesUtils.java
+++ b/core/src/main/java/org/commonjava/maven/ext/manip/util/PropertiesUtils.java
@@ -205,6 +205,10 @@ public final class PropertiesUtils
                                          String newVersion, Object originalType )
                     throws ManipulationException
     {
+        if (oldVersion == null) {
+            return false;
+        }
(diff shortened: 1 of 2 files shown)


rnc:
Was this due to a null pointer from pluginalignment? I think the boolean check in PluginManipulator needs fixing instead?

rnc:
Alternate fix in #279

core/src/main/java/org/commonjava/maven/ext/manip/util/PropertiesUtils.java
```diff
@@ -205,6 +205,10 @@ public static boolean cacheProperty( Map<String, String> versionPropertyUpdateMa
                                          String newVersion, Object originalType )
                     throws ManipulationException
     {
+        if (oldVersion == null) {
+            return false;
+        }
+
         boolean result = false;
```
rnc: Should we inline the comparison e.g.
```if ( oldVersion != null && oldVersion.startsWith....```
so there is only one return path.

PR #279, opened 2015-11-20
Add test and fix for empty version in pluginMgmt

rnc:


pgier:
I verified that this fix also works for my test build.  What do you think about also taking my commit from 278?  The two commits don't conflict, and it seems like it would still be good to check for null "oldVersion" in PropertyUtils#cacheProperty.

rnc:
@pgier @jdcasey Copying here for posterity: In at least one case, if it was not handled correctly prior to that being call, it would fall over soon after. There is probably no great harm in doing it ... possibly checkStrictValue needs checks as well then.

PR #294, opened 2015-11-27
Add log file parameter to Cli tool. Don't fork Cli invocations. Remov…

rnc:
…e unused code. Prevent integration-tests from creating empty main jar and create a test jar instead. Avoid running setup every time (for when debugging).

### DIFF
--- a/cli/pom.xml
+++ b/cli/pom.xml
@@ -119,6 +119,10 @@
       <artifactId>logback-classic</artifactId>
     </dependency>
     <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
       <scope>compile</scope>
@@ -140,6 +144,21 @@
       </testResource>
     </testResources>
     <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <configuration>
+              <classifier>minimal</classifier>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-assembly-plugin</artifactId>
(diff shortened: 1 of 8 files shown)


cli/pom.xml
```diff
@@ -141,6 +145,21 @@
...
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <configuration>
+              <classifier>minimal</classifier>
+            </configuration>
```
rnc: @jdcasey : Due to the standard jar embedded logging dependencies I needed a standard jar for use in the integration tests. Naming might not be ideal....

cli/src/main/java/org/commonjava/maven/ext/manip/Cli.java
```diff
@@ -184,65 +188,85 @@ private void run( String[] args )
...
+            fileAppender.setContext( loggerContext );
+            fileAppender.setName( "fileLogging" );
+            fileAppender.setAppend( false );
+            fileAppender.setFile( cmd.getOptionValue( "l" ) );
+            fileAppender.start();
+
+            root.addAppender( fileAppender );
+            root.setLevel( Level.INFO );
+        }
         // Set debug logging after session creation else we get the log filled with Plexus
```
rnc: Added log file support to CLI

integration-test/src/test/java/org/commonjava/maven/ext/manip/CliTestUtils.java
```diff
@@ -119,99 +124,46 @@ public static void runLikeInvoker( String workingDir, String restURL )
...
-            String.format( "%s -jar %s/pom-manipulation-cli.jar %s %s", getJavaBin().getAbsolutePath(), BUILD_DIR, stringParams, stringArgs );
+        logger.info( "Invoking CLI with {} ", arguments );
+        int result = new Cli().run( arguments.toArray( new String[arguments.size()] ) );
 
-        return runCommandAndWait( command, workingDir, null );
-    }
+        // This is a bit of a hack. The CLI, if log-to-file is enabled resets the logging. As we don't fork and run
+        // in the same process this means we need to reset it back again. The benefit of not forking is a simpler test
+        // harness and it saves time when running the tests.
+        final ch.qos.logback.classic.Logger root = (ch.qos.logback.classic.Logger) LoggerFactory.getLogger( org.slf4j.Logger.ROOT_LOGGER_NAME );
```
rnc: Avoid forking the CLI and just call inline. However, it does mean I need to reset the logging after....

PR #295, opened 2015-11-30
Skip install/deploy for integration test module.

jdcasey:


rnc:
@jdcasey Do we still need this?

PR #303, opened 2016-01-22
Handle errors from REST call better

rnc:
When the DA servers were being moved I was getting HTML 404/405 etc pages ... which then the  ProjectVersionRefMapper.java tried to parse and fell over on.


rnc:
retest this please

PR #313, opened 2016-03-06
squid:S1192 - String literals should not be duplicated

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S1192 - String literals should not be duplicated

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S1192

Please let me know if you have any questions.

M-Ezzat


jdcasey:
Sorry, I'm going to close this. The only really valid uses of the "don't repeat string literals" here are the 'target' and 'printGAVTC' strings. The others are log messages and moving their values to a constant IMO makes it harder to understand just by reading the code. The decrease in readability outweighs the minimal increase in performance or resilience.

Issue #328, opened 2016-04-01
Extend to handle GAV relocation in the xml file for maven-assembly-plugin


paulgallagher75:
The relocation code that was added had the knock on effect that the xml file for the maven-assembly-plugin does not get relocated and therefore the distro contains the wrong artifacts e.g. (activemq-artemis for AMQ7)

The top level pom.xml -
```
         <dependency>
            <groupId>org.apache.tomcat</groupId>
            <artifactId>tomcat-servlet-api</artifactId>
            <version>8.0.23</version>
         </dependency>
```
BECOMES
```
         <dependency>
            <groupId>org.jboss.spec.javax.servlet</groupId>
            <artifactId>jboss-servlet-api_3.1_spec</artifactId>
            <version>1.0.0.Final-redhat-1</version>
         </dependency>
```
However the artemis-distribution/src/main/assembly/dep.xml which is used to drive the maven-assembly-plugin has -
```
            <include>org.apache.tomcat:tomcat-servlet-api</include>
```
When it should be changed to -
```
            <include>org.jboss.spec.javax.servlet:jboss-servlet-api_3.1_spec</include>
```

rnc:
Currently PME only examines XML that are in the maven model not arbitrary XML files used by the assembly plugin. Also I believe the assembly plugin can use a slightly different format i.e. it doesn't specify version just group:artifact (and optionally type, classifier). 
A workaround would be to use the groovy plugin. It might be  possible to specify arbitrary extra files for PME to work on .... but what happens if they are not XML ?

paulgallagher75:
The problem is that the dep.xml file is not part of the PME model that gets passed to the groovy script (I think, please correct if I am wrong), and when I added in 'groovy.utils.XmlParser or groovy.utils.XmlSlupper', they were not in the classpath when the groovy script runs under PME, as I don't know how to add arbitrary dependencies at runtime for PME I resorted to just using a patch file to change it.

I think this may have brought up the problem of runtime dependencies for PME, I think we need a way to config that through the meta info of the cfg files.

jdcasey:
this is starting to sound more like we would need a "main" command for pom.xml files and another type (or multiple types?) of command that we could add to the PME execution to handle extra files. Assembly XMLs are definitely not going to the end of this...and at that point it's not a huge leap into processing Ant or even Gradle files.

[...]

PR #345, opened 2016-04-18
Issue #328 Add XML Manipulator

rnc:


core/src/test/java/org/commonjava/maven/ext/manip/impl/DistributionEnforcingManipulatorTest.java
```diff
@@ -366,7 +366,7 @@ public void before()
 
         final GalleyAPIWrapper wrapper = new GalleyAPIWrapper( galleyInfra );
 
-        manipulator = new DistributionEnforcingManipulator( wrapper );
+        manipulator = new DistributionEnforcingManipulator( );
+        FieldUtils.writeField (manipulator, "galleyWrapper", wrapper, true);
     }
```
jdcasey: What's the advantage of doing it this way? Doesn't this risk something breaking in weird ways if the class itself cannot specify the minimum parameter set for construction?

core/src/test/java/org/commonjava/maven/ext/manip/io/ModelResolverTest.java
```diff
@@ -52,9 +51,7 @@ public void resolveArtifactTest()
                             "cache-dir" ) );
         final GalleyAPIWrapper wrapper = new GalleyAPIWrapper( galleyInfra );
         final ModelIO model = new ModelIO();
-        final Field f = ModelIO.class.getDeclaredField( "galleyWrapper" );
-        f.setAccessible( true );
-        f.set(model, wrapper);
+        FieldUtils.writeField( model, "galleyWrapper", wrapper, true );
```
jdcasey: Again, why not create a constructor in ModelIO that allows it to specify the minimal components it needs? I see how FieldUtils is nicer than straight reflection, but it's still reflection underneath it all...

rnc: Discussed offline with @jdcasey.

Adding a specific constructor means we also needed to create a default constructor for the plexus injection. I was trying to minimise accessors / improve encapsulation in the main code base. However reflection isn't _that_ nice either ;-) Alternatives might be to make the field package accessible and initialise it on creation or create a package accessible setter to use. And another solution is to use Mockitto `@InjectMocks` / `@Spy` which allows a real object to be injected in and keep the cleaner 'private' fields/ctors.

jdcasey: Yeah, I think using a library we already have (commons-lang / FieldUtils) is simpler here, since the mocking library will be using this same technique under the covers anyway.

PR #348, opened 2016-04-20
PME removes SNAPSHOT from the version but does not use latest increme…

jsenko:
…ntal suffix.

IRC log:
reporter: I have sources in SNAPSHOT version: g:a:1.0-SNAPSHOT,
 when building [...] these sources, PME changes the version to 1.0.0-release-1.
reporter: When I do rebuild [...], PME changes the version again to 1.0.0-release-1.
reporter: Should it be rather changed to 1.0.0-release-2?
reporter: I think there is problem, because PME asks [the REST service] about g:a:1.0-SNAPSHOT and
 [it] doesn't find the best-match version. I think [...] PME should strip -SNAPSHOT before asking [...].


jdcasey:
add to whitelist

rnc:
@jsenko Thanks! Should we have a test for this?

jdcasey:
@jsenko +1 Thanks! I would love to see a test too, so we don't get regressions on this point later.

rnc:
#351 is an alternative implementation for this.

rnc:
@jsenko Apologies but due to timing we need to make a new PME release for @paulgallagher75 therefore I am closing this in favour #351

Issue #347, opened 2016-04-19
Add support for log-context to the PME CLI


rnc:
_From internal ticket NCL-1846_

PME needs to support receiving the log-context ID string from other tools (possibly as a parameter, but environment variable seems more akin to HTTP header) and sending it to DA via the log-context header.

rnc:
Fixed by #349

PR #349, opened 2016-04-20
Issue #347 / NCL-1846 Add LogContext facility

rnc:


rnc:
@Aszc : check this will be ok with repour ?

ASzc:
The code appears to do what it should, on both inspection and trial-execution. Once Repour adopts a version of PME with `--log-context` available, I will add it to the standard command line.

PR #357, opened 2016-05-03
NCL-1404 Add reading from local properties/yaml file

rnc:


io/src/main/java/org/commonjava/maven/ext/manip/io/ConfigIO.java
```diff
@@ -0,0 +1,116 @@
...
+
+@Component( role = ConfigIO.class )
+public class ConfigIO
+{
+    private final Logger logger = LoggerFactory.getLogger( ConfigIO.class );
+
+    private final static String propertyFileString = "pme.properties";
+    private final static String yamlPMEFileString = "pme.yaml";
+    private final static String yamlPNCFileString = "pnc.yaml";
+
```
jdcasey: Why do we have two different YAML files supported?

io/src/test/resources/org/commonjava/maven/ext/manip/io/pme.yaml
```diff
@@ -0,0 +1,11 @@
...
+  configKey: value
+  configHierarchy:
+    configKey1: value1
+
+pme:
+  version.suffix: redhat-1
+  dependencyExclusion.com.google.guava:guava@*: 18.0
+  repo-reporting-removal: true
+  jsonUpdate: >
+    distro/data/src/main/resources/data/basic-settings.apiman.json:$$..apimanVersion:${apiman:version}\,manager/ui/war/npm-shrinkwrap.json:$$..resolved:
```
jdcasey: are we going to get into trouble with escape characters, here or in the properties file? If we have an escape character we normally use on the CLI, and presumably YAML has one (I know properties files use '\' for escaping), how will those interact? Will we have things cropping up like: '\\\\\\something'?

io/src/main/java/org/commonjava/maven/ext/manip/io/ConfigIO.java
rnc: Originally it was `pme.yaml` but given its usage context in PNC I thought being able to name it `pnc.yaml` might be helpful. No difference in format ... just the naming support.

jdcasey:
Other than those concerns, I'm happy with this. Though I do also wonder how we're going to escape the chicken-egg scenario we've talked about in the past...I thought that had to do with the question of where (in Git) we would actually commit such a config file?

jdcasey:
Oh, and also...I cringe to bring this up...do we want to make the config file location itself configurable? If we did that, might we be able to load the config file from some external location (not the project Git repo)?

io/src/test/resources/org/commonjava/maven/ext/manip/io/pme.yaml
rnc: In terms of escaping anything for yaml, we don't do anything special (currently) in PME - we delegate everything to https://bitbucket.org/asomov/snakeyaml 
I suspect there still may be some corner cases come up ....

rnc:
@jdcasey Regarding committing the configuration file : currently the requirement on NCL-1404 was that the config file would be committed inside the productisation copy of the project repo. If we were to load the config file from a remote location - (e.g. another git repo) then there would still need to be way of tying this PME execution to that 'special case' configuration file.

PR #368, opened 2016-05-20
Improve property handling

rnc:
@jdcasey : Can you review this please :-)


core/src/main/java/org/commonjava/maven/ext/manip/util/PropertiesUtils.java
```diff
@@ -295,45 +347,21 @@ else if ( "project.version".equals( oldProperty ) )
...
+    public static String resolveProperties( List<Project> projects, String value )
                     throws ManipulationException
     {
-        String result = version;
+        final Properties amalgamated = new Properties();
 
-        if (version.startsWith( "${" ) )
+        // Reverse order so execution root ends up overwriting everything.
+        for ( int i = projects.size() - 1; i >= 0; i-- )
         {
```
jdcasey: nice!

core/src/test/java/org/commonjava/maven/ext/manip/util/PropertiesUtilsTest.java
```diff
@@ -15,33 +15,219 @@
...
+        session.setState( new VersioningState( p ) );
+
+        return session;
+    }
+
+    @Test
+    public void testUpdateNestedProperties() throws Exception
+    {
+        final Model modelParent = resolveRemoteModel( "org.infinispan:infinispan-bom:8.2.0.Final" );
+        Project pP = new Project( modelParent );
```
jdcasey: Is there any chance of just adding these to src/test/resources to avoid test failures due to network conditions? That makes it easier to work on PME while you're on a train, too... ;-D

core/src/test/java/org/commonjava/maven/ext/manip/util/PropertyInterpolatorTest.java
```diff
@@ -0,0 +1,106 @@
...
+            interp += pi.interp( d.getGroupId().equals( "${project.groupId}" ) ? project.getGroupId() : d.getGroupId() ) + ":" +
+                                pi.interp( d.getArtifactId().equals( "${project.artifactId}" ) ? project.getArtifactId() : d.getArtifactId() ) + System.lineSeparator();
+
+        }
+        assertTrue ( nonInterp.contains( "${" ) );
+        assertFalse ( interp.contains( "${" ) );
+    }
+
+    private Model resolveRemoteModel( final String resourceName ) throws Exception
+    {
```
jdcasey: Looks like this is used a few times...is it worth making a utils class (stateless) or a test fixture (stateful)?

jdcasey:
My few nit-picky comments aside, this looks good to me. I would kind of like to avoid remote dependencies for tests, though, if we can manage it without too much trouble.

PR #372, opened 2016-06-02
Improve project.version property handling

rnc:


rnc:
@jdcasey I realised that ProjectVersioningManipulator was changing project.version (which should be left to the ProjectVersionEnforcingManipulator). Detecting that meant I needed to change how the update code returned its status. Finally, added some more tests.

PR #400, opened 2016-06-20
Fault-tolerant version translator REST calls.

jsenko:
Split the big request into increasingly smaller parts when it fails
in an attempt to provide more stability against the translation backend.


jdcasey:
add to whitelist

rnc:
Thanks for the PR! Comments inline...

core/src/main/java/org/commonjava/maven/ext/manip/state/RESTState.java
```diff
@@ -33,11 +33,8 @@
     public RESTState( final Properties userProps )
     {
         restURL = userProps.getProperty( "restURL" );
-        /*
-      Limit on maximum count to make in the REST call.
-     */
-        int maxRestSize = Integer.parseInt( userProps.getProperty( "restMaxSize", "0" ) );
```
rnc: If we're removing restMaxSize the documentation in gh-pages will need updating (normally a separate PR).

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -1,12 +1,12 @@
...
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
  * Unless required by applicable law or agreed to in writing, software
```
rnc: Can we maintain the same license format as other files as otherwise the licensing check will fail

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -19,116 +19,172 @@
...
 import org.commonjava.maven.ext.manip.rest.mapper.ProjectVersionRefMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
```
rnc: No wildcard imports please ;-)

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -19,116 +19,172 @@
...
 
 /**
  * @author vdedik@redhat.com
+ * @author Jakub Senko <jsenko@redhat.com>
  */
 public class DefaultVersionTranslator
-    implements VersionTranslator
-{
+        implements VersionTranslator {
     private static final Random RANDOM = new Random();
```
rnc: There is an existing code style which should be importable into current IDEs. I think it would be preferable to maintain the same codestyle as the rest of the codebase.

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -19,116 +19,172 @@
...
-import java.util.Map;
-import java.util.Random;
+import java.util.*;
 
 import static org.apache.commons.lang.StringUtils.isNotEmpty;
 
 /**
  * @author vdedik@redhat.com
+ * @author Jakub Senko <jsenko@redhat.com>
  */
```
rnc: Fails to compile under JDK8

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -19,116 +19,172 @@
...
-     */
-    private final int maxRestSize;
+    public DefaultVersionTranslator(String endpointUrl) {
+        this(endpointUrl, 4);
+    }
 
-    public DefaultVersionTranslator( String endpointUrl, int maxRestSize )
-    {
+    public DefaultVersionTranslator(String endpointUrl, int chunkSplitCount) {
         this.endpointUrl = endpointUrl;
```
rnc: Do we need both ctors now?

io/src/main/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslator.java
```diff
@@ -19,116 +19,172 @@
...
+
+        while (!queue.isEmpty()) {
+            Task task = queue.remove();
+            task.executeTranslate();
+            if (task.isSuccess()) {
+                result.putAll(task.getResult());
+            } else {
+                if (task.canSplit()) {
+                    logger.debug("Failed to translate versions, splitting and retrying. " +
+                            "Chunk size is " + task.getChunkSize());
```
rnc: Is this going to output the previous chunk size rather than what it split it into?

io/src/test/java/org/commonjava/maven/ext/manip/rest/DefaultVersionTranslatorTest.java
```diff
@@ -165,22 +153,6 @@ public void testTranslateVersionsPerformance()
         versionTranslator.translateVersions( aLotOfGavs );
     }
 
-    private static final int restSize = 250;
-    @Test
-    @BMRule( name = "check-size",
-                    helper = "org.commonjava.maven.ext.manip.rest.HttpRequestWithBodyBytemanHelper",
```
rnc: I suspect if this test is no longer relevant we can remove the Helper class.

PR #410, opened 2016-06-27
NCL-2079 Change return codes for different failure types

rnc:
@matejonnet Most of PME already split out system error to manipulation errors.


matejonnet:
@rnc I don't see any error messages like "unable to run PME due to variable use in different versions." ?

PR #411, opened 2016-06-28
Cleanup - squid:S1312 - Loggers should be "private static final" and …

m-ezzat:
This pull request is focused on resolving occurrences of Sonar rule squid:S1312 - Loggers should be "private static final" and should share a naming convention

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid:S1312

Please let me know if you have any questions.

M-Ezzat


jdcasey:
@m-ezzat Actually, I've been slowly refactoring the code to use logger variables that are declared in the method itself. There's no real reason to store a class-level field for these, and certainly not a static variable. IMO it doesn't really have an advantage, and causes some marginal bloat since the variables cannot be garbage collected.

PR #418, opened 2016-07-21
Improve REST Completion status. Avoid aligning dependencies in first pass where we have an explicit override

rnc:


rnc:
@jdcasey This avoids the problem in PNC where DA returns a set of different versions for e.g.
```
            <dependency>
                <groupId>org.osgi</groupId>
                <artifactId>org.osgi.core</artifactId>
                <version>${osgi.version}</version>
            </dependency>
            <dependency>
                <groupId>org.osgi</groupId>
                <artifactId>org.osgi.enterprise</artifactId>
                <version>${osgi.version}</version>
            </dependency>
```
and then we get a property replacement clash when they are applied. By taking note of explicit overrides which are normally applied in a secondary pass this avoids the clash problem

PR #422, opened 2016-08-22
NCL-2188 Provide execution root GAV data for further processing.

jsenko:
A way that PME can provide some project data (execution root GAV) in a structured way to other tools that will work with the data further.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
+++ b/core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
@@ -230,15 +230,6 @@ public class ManipulationManager
         return changed;
     }
 
-
-    private void applyAfterManipulations( List<Project> projects, ManipulationSession session )
-                    throws ManipulationException
-    {
-        for ( final Manipulator manipulator : orderedManipulators ) {
-            manipulator.afterApplyChanges( projects, session );
-        }
-    }
-
     /**
      * After the modifications are applied, it may be useful for manipulators
      * to provide caller with a structured, computer-readable output or summary of the changes.
--- a/core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
+++ b/core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
@@ -15,6 +15,14 @@
  */
 package org.commonjava.maven.ext.manip;
 
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.introspect.VisibilityChecker;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.project.ProjectBuilder;
 import org.codehaus.plexus.component.annotations.Component;
@@ -23,11 +31,13 @@ import org.commonjava.maven.ext.manip.impl.Manipulator;
 import org.commonjava.maven.ext.manip.io.PomIO;
 import org.commonjava.maven.ext.manip.model.Project;
 import org.commonjava.maven.ext.manip.resolver.ExtensionInfrastructure;
+import org.commonjava.maven.ext.manip.state.State;
 import org.commonjava.maven.ext.manip.util.ManipulatorPriorityComparator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -58,6 +68,8 @@ public class ManipulationManager
     private static final String MARKER_PATH = "target";
 
     static final String MARKER_FILE =  MARKER_PATH + File.separatorChar + "pom-manip-ext-marker.txt";
+
+    static final String RESULT_FILE = MARKER_PATH + File.separatorChar + "pom-manip-ext-result.json";
 
     private final Logger logger = LoggerFactory.getLogger( getClass() );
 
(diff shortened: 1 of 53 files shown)


jsenko:
@rnc hi, could you take a look at this?

jdcasey:
add to whitelist

rnc:
@jsenko I'm on PTO but will look ASAP. Was there a reason for modifying PME rather than enhancing the build metadata plugin which already writes out relevant information?

jdcasey:
retest this please

jsenko:
@rnc @jdcasey hi, could you please decide on these changes? This solves an issue soon to become critical, so I have to know in advance if I have to find another way to solve this (you mentioned metadata plugin, could you elaborate? the plugin would have to run separately from PME right?).
I'm also not sure what causes the test failures. Any thoughts?

rnc:
@jdcasey Have emailed jsenko to ask "what information is required to be passed through for subsequent processing and when this information is required - after PME runs, after a build within 
PNC or...?"

jdcasey:
retest this please

core/src/main/java/org/commonjava/maven/ext/manip/impl/InfoManipulator.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override public void afterApplyChanges( List<Project> projects, ManipulationSession session )
+                    throws ManipulationException
+    {
+        Project executionRoot = getExecutionRoot( projects );
+        // Have to do this again to get modified values:/
+        final List<Project> projectsModified = pomIO.parseProject( executionRoot.getPom() );
+        executionRoot = getExecutionRoot( projectsModified );
+
```
rnc: I don't think this is required ; the Project instance contains the Model undergoing modification. Therefore that already has the correct GAV. 
Do we even need a new manipulator - the VersioningManipulator could fill in the InfoState once its completed altering the project?

core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
```diff
@@ -150,6 +156,32 @@ public void scanAndApply( final ManipulationSession session )
...
+
+        try
+        {
+            File resultFile = new File( session.getTargetDir(), ManipulationManager.RESULT_FILE );
+            boolean recreated = true;
+            resultFile.getParentFile().mkdirs();
+            if ( resultFile.exists() )
+            {
+                recreated = recreated && resultFile.delete();
+            }
```
rnc: Won't filewriter overwrite making this deletion code redundant. If this code was part of the marker file block above it would also make the mkdirs redundant.

core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
```diff
@@ -59,6 +63,8 @@
 
     static final String MARKER_FILE =  MARKER_PATH + File.separatorChar + "pom-manip-ext-marker.txt";
 
+    static final String RESULT_FILE = "pom-manip-ext-result.txt";
+
```
rnc: If we're writing a json file, why not use `.json` as the suffix?

core/src/main/java/org/commonjava/maven/ext/manip/impl/InfoManipulator.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.List;
+
+/**
+ * {@link Manipulator} implementation that gathers various information about the project.
+ * The information is intended to be saved into {@link ManipulationManager#RESULT_FILE}
+ */
+@Component( role = Manipulator.class, hint = "info-manipulator" ) public class InfoManipulator
+                extends AbstractNoopManipulator
+{
+    @Requirement private PomIO pomIO;
```
rnc: If you're providing an afterManip functionality then why do we need another Manipulator - shouldn't the ProjectVersioningManip be responsible for storing the new GAV once its completed altering it after manipulation?

core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
```diff
@@ -210,4 +243,53 @@ private void scan( final List<Project> projects, final ManipulationSession sessi
...
+        {
+            JsonNode node = MAPPER.convertValue( stateEntry.getValue(), JsonNode.class );
+            if ( node.isObject() && node.size() != 0 )
+            {
+                root.set( stateEntry.getKey().getSimpleName(), node );
+            }
+        }
+
+        return MAPPER.writeValueAsString( root );
+    }
```
rnc: Shouldn't we have a unit or integration test to verify this new functionality?

rnc:
@jdcasey @jsenko I can get the tests to pass with a slightly modified codebase (pretty much incorporating most of my above comments _except_ the lack of unit tests) as per https://github.com/rnc/pom-manipulation-ext/commit/f62841a3205f5f6dfe730b51e824bc623c2e5c7d

jsenko:
working on the test

core/src/main/java/org/commonjava/maven/ext/manip/ManipulationManager.java
```diff
@@ -210,4 +230,53 @@ private void scan( final List<Project> projects, final ManipulationSession sessi
...
 
+
+    private void applyAfterManipulations( List<Project> projects, ManipulationSession session )
+                    throws ManipulationException
+    {
+        for ( final Manipulator manipulator : orderedManipulators ) {
+            manipulator.afterApplyChanges( projects, session );
+        }
+    }
+
```
rnc: I don't think this is needed?

core/src/main/java/org/commonjava/maven/ext/manip/impl/AbstractNoopManipulator.java
```diff
@@ -0,0 +1,63 @@
...
+import static java.util.Collections.emptySet;
+
+/**
+ * Partial implementation of Manipulator interface with no-operation
+ * method bodies, to avoid unnecessary empty methods in child manipulators
+ * when that particular operation is not necessary.
+ */
+public abstract class AbstractNoopManipulator
+                implements Manipulator
+{
```
rnc: I don't think the overall change needs this addition?

core/src/main/java/org/commonjava/maven/ext/manip/impl/GroovyManipulator.java
```diff
@@ -161,6 +152,6 @@ public void init( final ManipulationSession session )
     @Override
     public int getExecutionIndex()
     {
-        return 99;
+        return 98;
     }
```
rnc: Can this be reverted please?

jsenko:
Done, but i think the noop abstr. class is useful, since there are a lot of noop methods in manipulators.

PR #435, opened 2016-10-18
Add alias dependencyOverride for dependencyExclusion

rnc:


core/src/main/java/org/commonjava/maven/ext/manip/state/DependencyState.java
```diff
@@ -96,7 +101,13 @@ public DependencyState( final Properties userProps )
         ignoreSuffix = Boolean.valueOf( userProps.getProperty( STRICT_ALIGNMENT_IGNORE_SUFFIX, "false" ) );
         failOnStrictViolation = Boolean.valueOf( userProps.getProperty( STRICT_VIOLATION_FAILS, "false" ) );
         remoteBOMdepMgmt = IdUtils.parseGAVs( userProps.getProperty( DEPENDENCY_MANAGEMENT_POM_PROPERTY ) );
+
         dependencyExclusions = PropertiesUtils.getPropertiesByPrefix( userProps, DEPENDENCY_EXCLUSION_PREFIX );
+        // If dependencyExclusions have not been set via exclusion prefix attempt to set it via the override alias prefix.
+        if ( dependencyExclusions.isEmpty())
+        {
+            dependencyExclusions = PropertiesUtils.getPropertiesByPrefix( userProps, DEPENDENCY_OVERRIDE_PREFIX );
```
pgier: Should this be additive instead of using one or the other?  What happens if the user uses both dependencyExclusions and dependencyOverrides?

core/src/main/java/org/commonjava/maven/ext/manip/state/DependencyState.java
```diff
@@ -41,6 +41,11 @@
     private static final String DEPENDENCY_EXCLUSION_PREFIX = "dependencyExclusion.";
 
     /**
+     * Merely an alias for {@link DependencyState#DEPENDENCY_EXCLUSION_PREFIX}
+     */
+    private static final String DEPENDENCY_OVERRIDE_PREFIX = "dependencyOverride.";
+
+    /**
```
pgier: I'd kind of like to have it the other way around, so "dependencyExclusion" is an alias for "dependencyOverride" even though it doesn't affect the functionality.  Based on the fact that the docs repeatedly use the term "override" in the feature descriptions, it seems like "dependencyOverride" should be the recommended parameter.

core/src/main/java/org/commonjava/maven/ext/manip/state/DependencyState.java
rnc: Thats a good question. Should it be additive, should one take precedence, should only one be allowed etc. Using `putAll` is one option to combine but is does mean we need to decide which should take precedence.

djcoleman: If dependencyOverride and dependencyExclusion are aliases, then I would expect them to be used interchangeably. If there are clashes between them, then I would flag that as an error, or some explicit overridable warning if that is indeed intended, so as to avoid any confusion about which one wins, or doing it by accident.

rnc:
Regarding your comment re site documentation - yes there will be, I was just getting this finalised first :-)

core/src/main/java/org/commonjava/maven/ext/manip/state/DependencyState.java
rnc: @djcoleman  That is a good point. Currently on the command line, later overrides earlier. However we can't determine the order between `-DdepEx1 -DdepOver1 -DdepEx2 -DdepOver4` so another option is to merge and flag an error if there is a conflict.

pgier:
The update looks good to me

PR #453, opened 2017-01-26
Fix for dependencyOverride for a child property but a top level prope…

rnc:
…rty matches that override


jdcasey:
retest this please

PR #488, opened 2017-04-14
Refactor version modifications to be stateless

pgier:
The Version class no longer stores any state about the version,
each version string modification is an independent operation which should
hopefully mean easier testing and debugging


pgier:
This is still a WIP at the moment, just want to do a bit of additional testing, so don't merge yet.

pgier:
I think this is ready to merge, @rnc can you take a look?
Hopefully, this refactor makes the code easier to follow and easier to fix bugs in the future.

rnc:
@pgier Nice! Looks good. Thanks :-)

@jdcasey @jbartece I may be remembering incorrectly but I thought at one point PNC / DA (?) was using the Version functionality from PME - this may affect it ?

jbartece:
Yes, AFAIK it's used at least in DA. @janinko ?

janinko:
In DA we use the Version for this:
```java
    private String toOsgi(String version) {
        return (new Version(version)).getOSGiVersionString();
    }
```

pgier:
Yes, "Version.getOsgiVersion(version)" should work fine as a replacement.

janinko:
Then I'm cool with the change :) @rnc

PR #515, opened 2017-09-23
PRODTASKS-210 Add inheritance to Project. Handle properties within groupId/artifactId and improve property resolving. 

rnc:
@jdcasey I don't know if you want to take a look at this as its one of the larger change sets of late :-) Happy to discuss offline.


rnc:
( discussed offline with John )

PR #628, opened 2019-04-10
Allow dependencyOverrides to reference a remote BOM as a version

ASzc:
This allows for more than one `dependencyManagement` BOM to be used when they
would otherwise conflict. The `dependencyOverride`/`dependencyExclusion`s limit
these "extra BOMs" effect to their requested modules/artifacts only.

The core RH-SSO server needs to align to EAP 7.x, but it also must build an
adapter that compiles against EAP 6.x. Before this change, the way to do this
would be:

```
-DdependencyManagement=org.jboss.eap:jboss-eap-parent:7.2.0.GA-redhat
-DdependencyRelocations.org.wildfly:@org.jboss.eap:=7.2.0.GA-redhat
-DdependencyOverride.org.jboss.as:*@*=7.5.20.Final-redhat-1
-DdependencyOverride.org.jboss.web:jbossweb@*=7.5.28.Final-redhat-1
...
```

Which works, but requires manually synchronizing the dependencyOverrides. With
this change, this is how it'd look:

```
-DdependencyManagement=org.jboss.eap:jboss-eap-parent:7.2.0.GA-redhat
-DdependencyRelocations.org.wildfly:@org.jboss.eap:=7.2.0.GA-redhat
-DdependencyManagement.eap6=org.jboss.bom:eap6-supported-artifacts:6.4.21.GA
-DdependencyOverride.org.jboss.as:*@*=eap6
-DdependencyOverride.org.jboss.web:jbossweb@*=eap6
...
```

Which results in less churn and manual work.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
@@ -134,6 +134,9 @@ public class DependencyManipulator implements Manipulator
         final RESTState restState = session.getState( RESTState.class );
         final List<ProjectVersionRef> gavs = depState.getRemoteBOMDepMgmt();
 
+        final Map<String, ProjectVersionRef> extraGAVs = depState.getExtraBOMs();
+        final Map<String, Map<ProjectRef, String>> extraBOMOverrides = depState.getExtraBOMDepMgmts();
+
         // While in theory we are only mapping ProjectRef -> NewVersion if we store key as ProjectRef we can't then have
         // org.foo:foobar -> 1.2.0.redhat-2
         // org.foo:foobar -> 2.0.0.redhat-2
@@ -155,6 +158,13 @@ public class DependencyManipulator implements Manipulator
                 // for strict override. However, it is undefined if strict is not enabled.
                 bomOverrides.putAll( rBom );
             }
+        }
+
+        // Load extra BOMs into seperate maps for accessing later, when applying the dependencyExclusions.
+        for ( Map.Entry<String, ProjectVersionRef> entry : extraGAVs.entrySet() )
+        {
+            extraBOMOverrides.put( entry.getKey(),
+                                   effectiveModelBuilder.getRemoteDependencyVersionOverridesByProject( entry.getValue() ) );
         }
 
         if ( depState.getPrecedence() == DependencyPrecedence.BOM )
@@ -338,7 +348,8 @@ public class DependencyManipulator implements Manipulator
         {
             moduleOverrides = applyModuleVersionOverrides( projectGA,
                                                            dependencyState.getDependencyExclusions(),
-                                                           moduleOverrides, explicitOverrides );
+                                                           moduleOverrides, explicitOverrides,
+                                                           dependencyState.getExtraBOMDepMgmts() );
             logger.debug( "Module overrides are:\n{}", moduleOverrides );
             logger.debug( "Explicit overrides are:\n{}", explicitOverrides);
         }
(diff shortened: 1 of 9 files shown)


rnc:
ok to test

rnc:
@ASzc Thanks for the contribution! About to go on pto but will try to review asap.

ASzc:
Ok, thanks

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -793,6 +804,8 @@ else if ( strict && oldVersion.contains( "$" ) &&
                 final boolean isModuleWildcard = currentKey.endsWith( "@*" );
                 logger.debug( "Is wildcard? {} and in module wildcard mode? {} ", isModuleWildcard, aWildcardMode );
 
+                String artifactGA = null;
```
rnc: Given its usage shouldn't this be immediately converted to a ProjectRef on construction? (to avoid repeated conversions)

rnc:
This is a nice idea! Thanks very much!

Few comments/questions:
1. Does it avoid the churn because you are referencing a fixed BOM rather than a set of incrementing -redhat-n ?
2. Ideally I'd like a documentation update (on the gh-pages branch) along with this.
3. [ minor ] in future can we keep the same coding style please :-D

ASzc:
For 1, not exactly. The work and churn comes from having to manually call up each new BOM version and copy over the dependencyOverride versions from that, which are often unique to the component.

I can do 2.

For 3, I did this in vim, so I was manually trying to keep to the style, but I do see some areas where I slipped. I'll see if I can feed it through eclipse and update the PR.

rnc:
@ASzc  Ah no worries, I use IntelliJ :-)

rnc:
@ASzc 
1. So you still have to specify for each dependency group:artfact in the affected sub-project
```
-DdependencyOverride.org.jboss.as:*@*=eap6
```
 but now you don't need to worry about what specific redhat increment they are as its dealt with transparently?

2. Thanks

ASzc:
The delta in any one case could be in the redhat suffix, or other parts of the version. Typically it might be a redhat suffix change (I haven't looked recently), but we want the latest EAP 6's versions for the EAP 6 adapter module, regardless of what kind of change is there versus the POM. If the new version is incompatible, it'll show up during the continuous compilation and testing process we're following.

io/src/main/java/org/commonjava/maven/ext/io/ModelIO.java
```diff
@@ -186,6 +186,18 @@ public File resolveRawFile( final ArtifactRef ref )
         return versionOverrides;
     }
 
+    public Map<ProjectRef, String> getRemoteDependencyVersionOverridesByProject( final ProjectVersionRef ref )
+                    throws ManipulationException
+    {
+        final Map<ProjectRef, String> versionOverrides = new LinkedHashMap<>();
```
rnc: Assuming I understand correctly - you need to convert to index off ProjectRef as that will be compared in the override g:a section (?)  wouldn't something like be simpler ? :
```
        return getRemoteDependencyVersionOverrides(ref).entrySet().stream().collect(
                        Collectors.toMap( e -> e.getKey().asProjectRef(), Map.Entry::getValue, (key1, key2) -> {
                            // Avoid problem where a ProjectRef leads to duplicates when the ArtifactRef didn't
                            return key1;
                        } ) );
```

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
ASzc: For sure, although fixing this existing code repetition would have muddied the diff a bit.

io/src/main/java/org/commonjava/maven/ext/io/ModelIO.java
ASzc: That's right. I assume there's some reason the other override maps use ArtifactID keys, but I don't for this, and ProjectIDs allow a simple lookup rather than iterating through a map as a list as `DependencyManipulator.removeGA()` does.

That would be a more succinct conversion, but I haven't worked on Java 8+ code much, so I wrote it in the old lambda-less style.

ASzc:
Added a PR for gh-pages

io/src/main/java/org/commonjava/maven/ext/io/ModelIO.java
rnc: No worries , thanks for changing.

PR #638, opened 2019-05-19
Cleaned up multiple logging statements

orpiske:
Cleaned up multiple logging statements to prevent concatenating strings unnecessarily and implement logging guards in a few occurrences where not doing so would result in associated methods being called even if the debug level was beyond the minimum used during the execution.


rnc:
ok to test

rnc:
Thanks for the contribution ; will look after our F2F this week.

PR #639, opened 2019-05-19
Treat HTTP status 503 as a recoverable since the server may be back in a bit

orpiske:
This should improve the behavior when using DA servers that suffer from reliability issues and get restarted/bounced from time to time.

### DIFF
--- a/io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
+++ b/io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
@@ -71,7 +71,7 @@ public class DefaultTranslator
 
     private final ListingBlacklistMapper lbm;
 
-    private final int RETRY_WAIT_DURATION = 30;
+    private int retryDuration = 30;
 
     /**
      * @param endpointUrl is the URL to talk to.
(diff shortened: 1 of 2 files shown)


rnc:
ok to test

rnc:
Thanks for the contribution ; will look after our F2F this week.

rnc:
This would need to be discussed with the PNC team (cc @jbartece ) as it was requested in NCL-2774. With any changes to the implementation here we should consider whether to upgrade unirest. Also I think we should have tests for this.

rnc:
Also can you add tests for this please.

orpiske:
> This would need to be discussed with the PNC team (cc @jbartece ) as it was requested in NCL-2774. With any changes to the implementation here we should consider whether to upgrade unirest. Also I think we should have tests for this.

TBH, that is fine for me. The only thing I need is PME to stop failing when DA is down or unavailable. I have had several parts of my pipeline failing because it goes down, only to be back a few seconds later.

> Also can you add tests for this please.

Absolutely. I put a pause on this work while I handle an ongoing release and I expect to be back to this soon.

orpiske:
> Also can you add tests for this please.

Done. I reused one of the existing tests for VersionTranslator. 

Also, I can squash the commits if you want. Just let me know.

rnc:
Mentioned to @janinko and @jbartece  and they had no objections. 

Thanks for the contribution, will merge.

PR #640, opened 2019-05-21
Log to report

orpiske:
This modifies the code so that the information that is presented on the post-alignment report can also be saved to multiple files (per project). 

For example, using this snipped for a report: 

```
INFO o.c.m.e.common.callbacks.LogReporter - ------------------- project org.apache.karaf:karaf 
INFO o.c.m.e.common.callbacks.LogReporter - 	Project version : 4.2.0.fuse-740005 ---> 4.2.0.fuse-740005-orpiske-00001

INFO o.c.m.e.common.callbacks.LogReporter - 	Dependencies : junit:junit:jar:4.12 --> junit:junit:jar:4.12.0.redhat-003 
INFO o.c.m.e.common.callbacks.LogReporter - 	Managed dependencies : junit:junit:jar:4.12 --> junit:junit:jar:4.12.0.redhat-003 
INFO o.c.m.e.common.callbacks.LogReporter - 	Managed dependencies : org.apache.karaf.management:org.apache.karaf.management.server:jar:4.2.0.fuse-740005 --> org.apache.karaf.management:org.apache.karaf.management.server:jar:4.2.0.fuse-740005-orpiske-00001
``` 

If the option `--report-dir /home/opiske/tmp/alignment-report` is passed to PME, then it would create these files with the following content:

- karaf.version
`org.apache.karaf:karaf:4.2.0.fuse-740005=org.apache.karaf:karaf:4.2.0.fuse-740005-orpiske-00001`
- karaf.dependencies:
```Dependencies;junit:junit:jar:4.12;junit:junit:jar:4.12.0.redhat-003
Managed dependencies;junit:junit:jar:4.12;junit:junit:jar:4.12.0.redhat-003
Managed dependencies;org.apache.karaf.management:org.apache.karaf.management.server:jar:4.2.0.fuse-740005;org.apache.karaf.management:org.apache.karaf.management.server:jar:4.2.0.fuse-740005-orpiske-00001
```
- karaf.dependencies. It's usually empty, unless reportNonAligned property is set, but if given should print something similar to:
```
Dependencies;junit:junit:jar:4.12
```

So, why this is useful? Because we want to integrate PME in our pipeline, but we don't deploy the projects after they are aligned. Therefore, we cannot rely on the DA to update the overrides. By processing the contents of this report, we can process the data and feed into the subsequent projects on the pipeline and ensure they use the aligned dependencies. 

An additional benefit of this approach is that we currently we do that using maven-help-plugin, but it is very slow. With this report we can cut a lot of time on our pipeline.

### DIFF
--- a/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
+++ b/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
@@ -55,6 +55,8 @@ import org.codehaus.plexus.component.repository.exception.ComponentLookupExcepti
 import org.commonjava.maven.atlas.ident.ref.ArtifactRef;
 import org.commonjava.maven.atlas.ident.ref.InvalidRefException;
 import org.commonjava.maven.ext.common.ManipulationException;
+import org.commonjava.maven.ext.common.callbacks.ComparatorCallback;
+import org.commonjava.maven.ext.common.callbacks.FileReporter;
 import org.commonjava.maven.ext.common.model.SimpleScopedArtifactRef;
 import org.commonjava.maven.ext.core.ManipulationManager;
 import org.commonjava.maven.ext.core.ManipulationSession;
@@ -181,6 +183,11 @@ public class Cli
                                  .numberOfArgs( 2 )
                                  .desc( "XPath tester ( file : xpath )" )
                                  .build() );
+        options.addOption( Option.builder( "" )
+                .longOpt( "report-dir" )
+                .desc( "Creates a report with all the alignment changes and saves to the given file" )
+                .numberOfArgs( 1 )
+                .build() );
 
         CommandLineParser parser = new DefaultParser();
         CommandLine cmd;
@@ -394,6 +401,11 @@ public class Cli
             }
             else
             {
+                if (cmd.hasOption( "report-dir" )) {
+                    String dir = cmd.getOptionValue( "report-dir" );
+                    manipulationManager.getPostAlignmentCallbacks().add(new ComparatorCallback(new FileReporter(dir)));
+                }
+
                 manipulationManager.scanAndApply( session );
             }
         }
--- a/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
+++ b/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
@@ -183,7 +183,7 @@ public class Cli
                                  .numberOfArgs( 2 )
                                  .desc( "XPath tester ( file : xpath )" )
                                  .build() );
-        options.addOption( Option.builder( "" )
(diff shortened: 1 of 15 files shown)


orpiske:
Finally, another point about this is that I implemented the code in a way that it can be leveraged in the future. Should any user want to do something like store this changes on a database or whatever they seem fit, they can just implement the Report interface or extend the PrintableReport class.

rnc:
I have been thinking about this and it seems to tie very closely to the Fuse pipeline (which may not be suitably generic to apply to other products and as you know we have pointed out things like the osgi problem should have been solved differently). I think potentially a better solution may be to extend the existing json output and combine with the project comparator to put the differences in the json file.

orpiske:
I am not sure why do you think it ties closely to our pipeline. There's nothing on the code that binds specifically to the way we do. The only thing the code does is report what has been aligned (and, in fact, this is something [that other teams need as well](https://github.com/k-wall/alignment-reporter-plugin)).

I also disagree that putting the output in the json would be better: it's harder to read, process and compute. A text-based report is precisely what we need: something that goes straight to the point, is easy to read and process and doesn't over-complicate.

rnc:
I was unaware of that plugin ; Keith never mentioned he needed something (and to be fair he is in your product bundle, not in another team). For a lot of projects this is going to produce a huge number of files; its something I will have to have a think about. For instance I am not convinced it will help with scenarios where you have problems with dependencies within PNC.

orpiske:
> I was unaware of that plugin ; Keith never mentioned he needed something (and to be fair he is in your product bundle, not in another team). For a lot of projects this is going to produce a huge number of files; its something I will have to have a think about. For instance I am not convinced it will help with scenarios where you have problems with dependencies within PNC ; I realise you are changing your process significantly - I think perhaps looking at an option of not aligning already aligned dependencies might also be useful.

I understand the scale of the projects we have is, probably, bigger than the average, so I made the reports entirely optional. 
Since we are moving to have manual alignments ourselves, having these reports may make things easier to analyze.

rnc:
X-reference : Internal jira : PRODTASKS-676

rnc:
@orpiske As part of the work with Keith to provide a base for future design/integration for the concept of his plugin I have sent a PR for #667. This supplements the existing textual logging (and report-to-file ability) with JSON output and a set of POJOs for reuse.

As this point, given the existing code now has text, logging and potential future json ability I would like to close this PR in preference for that.

orpiske:
@rnc thanks. That's ok: I think our team has decided to take a different approach about how to handle dependency alignments and I think we won't need this feature anymore. In case this change in the future, I will rework this to use the plugin mechanism to be implemented. I will close the PR.

PR #641, opened 2019-05-21
Overrides from file

orpiske:
This one depends on PR #640 and adds a functionality that allows it to read the overrides from a text file. The reason for this feature is that we have a very big pipeline. After every build, we update the list of overrides and put it into a text file containing a list of overrides that is passed as input arguments to PME. The file looks like this:: 

```-DdependencyOverride.org.apache.karaf.webconsole:org.apache.karaf.webconsole.instance@*=4.2.0.fuse-740006-myorg-00001 
-DdependencyOverride.org.apache.karaf.webconsole:webconsole@*=4.2.0.fuse-740006-myorg-00001 
-DdependencyOverride.org.apache.karaf.wrapper:org.apache.karaf.wrapper.core@*=4.2.0.fuse-740006-myorg-00001 
```

However, since our project is significantly complex and the overrides file got so big that we cannot pass it as arguments since it exceeds the number of arguments that can be passed. 

Although we know about using a BOM for the overrides, we prefer to use a text file because it is simpler and faster to generate.

### DIFF
--- a/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
+++ b/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
@@ -76,6 +76,7 @@ import javax.xml.xpath.XPath;
 import javax.xml.xpath.XPathConstants;
 import javax.xml.xpath.XPathFactory;
 import java.io.File;
+import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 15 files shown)


rnc:
ok to test

rnc:
Can you explain why the existing configuration file support is not sufficient? ( https://release-engineering.github.io/pom-manipulation-ext/guide/configuration.html )

orpiske:
Because we are generating it automatically. It's cleaner to generate just a text file w/ the contents we need.

rnc:
I still am not clear why the existing properties configuration file support is not sufficient.

orpiske:
> I still am not clear why the existing properties configuration file support is not sufficient.

How should be the format of the properties file for this? Would this work?

`dependencyOverride.org.apache.karaf.webconsole:webconsole@*=4.2.0.fuse-740006-myorg-00001`

rnc:
I think it should do. Its just a standard java properties file. One thing to note though is that (currently) the naming / location is fixed ( https://release-engineering.github.io/pom-manipulation-ext/guide/configuration.html )

orpiske:
Ok. I am going to give a try with this format for the next version of our builder. Thanks!

rnc:
I am closing this for now but would be happy to take another look (or we could enhance the existing config file support)

PR #667, opened 2019-08-10
PRODTASKS-676 Add JSON Comparator output

rnc:


rnc:
Documentation in #668

PR #672, opened 2019-08-21
allow groovy scripts to remove projects

michalszynkiewicz:


rnc:
Thanks for the contribution. Am on PTO at the moment so will look in detail when I get back.

Are you able to supply a test case as well please?

rnc:
ok to test

rnc:
Thanks for the tests! Out of curiosity (although might be preferable to do on email) can you describe the use-case please?

michalszynkiewicz:
S.W.I.M. builds a project that has a bom file with a list of all extensions (and third parties).
During the standard alignment, all submodules are reversioned to internal versions (regardless of wether we want to build them or not).
Removing a project from the PME run (as in the test), along with some other groovy magic, results in a submodule not being. This way the BOM doesn't contain coordinates of non-existing artifacts.

Not sure it's clear what I meant from what I wrote :) 
I think I can do alignment locally for a while and we can discuss it when you're back from PTO.

rnc:
Sure. What is the priority of this (release timings eg waiting for pnc 2.0)

michalszynkiewicz:
Unless we find a different way to achieve what we need, it would have to go in before pnc 2.0.

PR #680, opened 2019-09-24
NCL-4779 Change versionIncrementalSuffixPadding default to 5. Fix iss…

rnc:
…ue with padding and static suffix


core/src/main/java/org/commonjava/maven/ext/core/impl/VersionCalculator.java
```diff
@@ -118,9 +118,11 @@ public VersionCalculator( final GalleyAPIWrapper readerWrapper )
...
             if ( buildNumber > 0 )
             {
-                String paddedBuildNum = StringUtils.leftPad( Integer.toString( buildNumber ),
-                                                             Version.getBuildNumberPadding( state.getIncrementalSerialSuffixPadding(),
-                                                                                            versionsWithBuildNums ), '0' );
+                // We ONLY pass the incrementalSerialSuffixPadding if we are using incrementalSuffix (i.e. not static suffix).
+                // This keeps it consistent with the 'calculate' method..
+                String paddedBuildNum = StringUtils.leftPad(
+                                Integer.toString( buildNumber ), Version.getBuildNumberPadding(
+                                                ( state.getSuffix() == null ? state.getIncrementalSerialSuffixPadding() : 0 ), versionsWithBuildNums ), '0' );
```
rnc: @vibe13 Turns out when changing the tests to use the new default it uncovered a bug within static suffix and padding (not incremental which is used in PNC).

rnc:
@vibe13 Also changed the tests to be a better mix of padding 0 / 5.

PR #693, opened 2019-10-08
PRODTASKS-698 Upgrade Unirest to v3

rnc:


vibe13:
@rnc it looks good to me, great work! I like the reuse of DA reports-model. Btw, I see you are mapping the whitelist information, and probably returning it back in the results? That information is not used anymore, so it might be misleading to users (whitelist tables are deprecated in DA now cc @janinko right)?

janinko:
@vibe13 @rnc the Whitelist in DA was never properly populated and will be replaced by new service in the future. So yeah, I would consider it deprecated.

PR #697, opened 2019-10-11
NCL-5256 Improve exception propagation

rnc:


core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -184,7 +186,15 @@ void applyGroovyScript( List<Project> projects, Project project, File groovyScri
...
             {
-                throw new ManipulationException( "Unable to parse script", e );
+                //noinspection ConstantConditions
+                if ( e instanceof ManipulationException )
+                {
+                    throw ((ManipulationException)e);
+                }
+                else
+                {
+                    throw new ManipulationException( "Unable to parse script", e );
```
rnc: @paulgallagher75 I wonder if this exception would be better phrased as `Problem running script` ?

PR #706, opened 2019-11-06
NCL-5299 Ensure that lines use platform-dependent end-of-line

dwalluck:


rnc:
@dwalluck  Is this causing a particular problem for you? I believe by default it uses the system linefeed. I suspect that changing just this may not be consistent with other parts of the code?

dwalluck:
> @dwalluck Is this causing a particular problem for you? I believe by default it uses the system linefeed. I suspect that changing just this may not be consistent with other parts of the code?

Without this PR, there is a test failure in `JSONIOTest` because the JSON file produced is not byte-for-byte compatible with the original. Replacing `FileUtils.contentEquals()` with `FileUtils.contentEqualsIgnoreEOL()` may also work for passing the test, but the problem still remains that we will produce different JSON files depending on platform and I am not sure that this is what we want.

There are really two separate issues here.

The parser doesn't know what the EOL in the original file is.  I suppose it could try to guess what the line-ending is in the original file, but this PR does not address that. Furthermore, if the rest of the file is not written with `'\n'` then https://github.com/release-engineering/pom-manipulation-ext/blob/2f99cd9d0b90fd15ecd0c5d43778bbfea9bd9408/io/src/main/java/org/commonjava/maven/ext/io/JSONIO.java#L81 should presumably write a `System.lineSeparator()` instead of `'\n'` or you end up with inconsistent EOLs within the same file.

This PR is to fix the test failure on platforms where EOL is not `'\n'` and to make it consistent with line 81. I'd certainly prefer if the JSON files produced by the `JSONIO` class do not differ based on what system it was run on as we don't get reproducible output across all platforms. In my opinion, I think having the output be reproducible is an important feature in itself regardless of whether we can solve the issue of trying to make the input and output files match.

dwalluck:
@rnc I think the options are to change the test to use `FileUtils.contentEqualsIgnoreEOL()` and also change line 81 to write `System.lineSeparator()`, or to change the output so that it always writes `'\n'` regardless of platform. I personally think that it makes more sense to force the newlines to be LF rather than defaulting to a system-dependent newline because I think that the tool should be reproducible across platforms.

By the way, Travis has built-in support for osx in addition to linux, and there is also a possibility of using another CI if you want to build on Windows as well. This could catch these types of issues before they happen.

dwalluck:
@rnc Maybe do the same thing as is done in `PomIO.determineEOL()`? I only see two non-test files referencing a line-separator: `PomIO.java` and `ProjectComparator.java`.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -308,7 +308,7 @@ else if ( new File( target.getParentFile(), ManipulationManager.MARKER_FILE ).ex
         {
             // Note : don't print out settings information earlier (like when we actually read it) as the logging
             // isn't setup then.
-            logger.debug( "Using local repository \n{} and found global settings file in {} with contents \n{} and user settings file in {} with contents \n{}",
+            logger.debug( "Using local repository %n{} and found global settings file in {} with contents %n{} and user settings file in {} with contents %n{}",
```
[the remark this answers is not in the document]
rnc: Hmmm. Just modified a test and got:
```
[main] DEBUG o.c.m.e.c.u.ProjectInheritanceTest - Got the following version suffixes:%n  Static: org.commonjava.maven.ext:pom-manipulation-parent:3.9-SNAPSHOT [pom=/home/rnc/Work/PME/pom-manipulation-ext/pom.xml]%n
```

core/src/main/java/org/commonjava/maven/ext/core/impl/DistributionEnforcingManipulator.java
```diff
@@ -302,7 +302,7 @@ private Xpp3Dom getConfigXml( final Node node )
         catch ( final XmlPullParserException | IOException e )
         {
             throw new ManipulationException(
-                                             "Failed to re-parse plugin configuration into Xpp3Dom: %s\nConfig was:\n%s",
+                                             "Failed to re-parse plugin configuration into Xpp3Dom: %s%nConfig was:%n%s",
```
rnc: Another %n?

io/src/main/java/org/commonjava/maven/ext/io/FileIO.java
```diff
@@ -16,23 +16,42 @@
...
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.UUID;
 
+import com.google.common.primitives.Chars;
```
rnc: Where is this from/used?

io/src/main/java/org/commonjava/maven/ext/io/ModelIO.java
```diff
@@ -109,7 +109,7 @@ public Model resolveRawModel( final ProjectVersionRef ref )
         }
         catch ( final TransferException e )
         {
-            throw new ManipulationException( "Failed to resolve POM: %s.\n--> %s", e, ref, e.getMessage() );
+            throw new ManipulationException( "Failed to resolve POM: %s.%n--> %s", e, ref, e.getMessage() );
```
rnc: More %n

io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java
```diff
@@ -56,7 +56,7 @@ public XMLIO()
 
             transformer = TransformerFactory.newInstance().newTransformer();
             transformer.setOutputProperty( OutputKeys.INDENT, "yes");
-            transformer.setOutputProperty( OutputKeys.ENCODING, StandardCharsets.UTF_8.toString());
+            transformer.setOutputProperty( OutputKeys.ENCODING, StandardCharsets.UTF_8.name() );
```
rnc: Why this change?

io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java
```diff
@@ -58,7 +58,7 @@ public void write( Settings settings, File settingsFile )
     {
         try
         {
-            PrintWriter printWriter = new PrintWriter( settingsFile, StandardCharsets.UTF_8.toString() );
+            PrintWriter printWriter = new PrintWriter( settingsFile, settings.getModelEncoding() );
```
rnc: Why this change please?

rnc:
I've added some more comments. As a slight aside, is it your IDE reformatting and changing code that isn't directly related to your change? It makes the diffs much larger and harder to parse. Is there any way of only changing what is directly relevant?

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -369,18 +375,18 @@ else if ( cmd.hasOption( 'p' ) || cmd.hasOption( "printGAVTC" ) )
                     {
                         if ( cmd.hasOption( "printGAVTC" ) )
                         {
-                            FileUtils.writeStringToFile( output, String.format( "%-80s\n", a), StandardCharsets.UTF_8, true );
+                            FileUtils.writeStringToFile( output, String.format( "%-80s%n", a ), StandardCharsets.UTF_8, true );
                         }
                         else
                         {
-                            FileUtils.writeStringToFile( output, a.asProjectVersionRef().toString() + '\n', StandardCharsets.UTF_8, true );
+                            FileUtils.writeStringToFile( output, String.format( "%s%n", a.asProjectVersionRef().toString() ), StandardCharsets.UTF_8, true );
```
rnc: Why use String.format rather than keeping the original and using lineSeparator?

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -768,10 +768,10 @@ else if ( strict && oldVersion.contains( "$" ) &&
     {
         final Map<ArtifactRef, String> remainingOverrides = new LinkedHashMap<>( originalOverrides );
 
-        if (logger.isDebugEnabled())
+        if ( logger.isDebugEnabled() )
         {
-            logger.debug("Calculating module-specific version overrides. Starting with:\n  {}",
-                    join(remainingOverrides.entrySet(), "\n  "));
+            logger.debug( "Calculating module-specific version overrides. Starting with:{}  {}", System.lineSeparator(),
+                    join( remainingOverrides.entrySet(), String.format( "%n  " ) ) );
```
rnc: As per above re String.format

core/src/test/java/org/commonjava/maven/ext/core/impl/ProjectVersionManipulatorTest.java
```diff
@@ -651,8 +651,8 @@ private TestVersioningModifier newVersioningModifier()
                 if ( applyVersioningChanges( new Project ( project.getOriginalModel()), state ) )
                 {
                     final String v = _versionsByGAV.get( SimpleProjectVersionRef.parse( gav( project ) ) );
-                    logger.info( project.getName() + " (" + gav( project ) + "): VERSION MODIFIED\n    New version: "
-                        + v );
+                    logger.info( "{} ({}): VERSION MODIFIED{}    New version: {}",
+                        project.getName(), gav( project ), System.lineSeparator(), v );
```
rnc: Given this part hasn't been changed since 2014/5 .... I'd be tempted just to remove the logger line. We don't need it in tests (and if we do have logging / `System.out.println` I tend to use them in conjunction with https://stefanbirkner.github.io/system-rules/#SystemErrAndOutRule and mute for success.

core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java
```diff
@@ -303,12 +303,12 @@ public void testIsSnapshot()
     @Test
     public void testIsEmpty()
     {
-        assertTrue( Version.isEmpty(null) );
-        assertTrue( Version.isEmpty("") );
-        assertTrue( Version.isEmpty("  \n") );
+        assertTrue( Version.isEmpty( null ) );
+        assertTrue( Version.isEmpty( "" ) );
+        assertTrue( Version.isEmpty( String.format( "  %n" ) ) );
```
rnc: Why not just use `"  " + System.lineSeparator()` ?

io/src/main/java/org/commonjava/maven/ext/io/CharsetAwareJacksonJsonProvider.java
```diff
@@ -0,0 +1,83 @@
...
+
+import org.apache.commons.io.ByteOrderMark;
+import org.apache.commons.io.input.BOMInputStream;
+
+import com.fasterxml.jackson.core.JsonEncoding;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.jayway.jsonpath.InvalidJsonException;
+import com.jayway.jsonpath.spi.json.JacksonJsonProvider;
+
+public class CharsetAwareJacksonJsonProvider extends JacksonJsonProvider
```
rnc: Nice! Can we add some javadoc for the class description please?

io/src/main/java/org/commonjava/maven/ext/io/JSONIO.java
```diff
@@ -30,35 +37,140 @@
...
+
+    private File jsonFile;
+
+    private JsonEncoding encoding;
+
+    private Charset charset;
+
+    private String eol;
+
+    private JsonEncoding detectEncoding( File jsonFile ) throws ManipulationException
```
rnc: Nice!

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
```diff
@@ -101,13 +100,14 @@
...
             Model raw;
+
             try ( InputStream in = new FileInputStream( pom ) )
             {
                 raw = new MavenXpp3Reader().read( in );
             }
             catch ( final IOException | XmlPullParserException e )
             {
-                throw new ManipulationException( "Failed to build model for POM: %s.\n--> %s", e, pom, e.getMessage() );
+                throw new ManipulationException( "Failed to build model for POM: %s.%n--> %s", e, pom, e.getMessage() );
```
rnc: More %n

io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java
rnc: Ah are you trying to get hold of 
```<?xml version="1.0" encoding="utf-8"?>``` ?

File: io/src/main/java/org/commonjava/maven/ext/io/PomIO.java (hunk 13 above)
Comment (dwalluck) [reply io/src/main/java/org/commonjava/maven/ext/io/PomIO.java]: It should be OK inside an exception message as I see it using other formatting codes in there. I will verify.

File: core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java (hunk 10 above)
Comment (dwalluck) [reply core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java]: It could, it just looked closer to the original that way.

**Comment by dwalluck** [comment]
> As a slight aside, is it your IDE reformatting and changing code that isn't directly related to your change? It makes the diffs much larger and harder to parse. Is there any way of only changing what is directly relevant?

Would it work on the pull request to click the icon at the top and select "Just for now
 Hide whitespace changes"?

File: io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java (hunk 5 above)
Comment (dwalluck) [reply io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java]: Because `name()` (Returns this charset's canonical name) should be what we want, not `toString()` (Returns a string describing this charset). If `name()` and `toString()` match that's more of a coincidence:

File: io/src/main/java/org/commonjava/maven/ext/io/FileIO.java (hunk 3 above)
Comment (dwalluck) [reply io/src/main/java/org/commonjava/maven/ext/io/FileIO.java]: This is just a case of the IDE importing something while typing (probably wanted `StandardCharsets`). I wish there was just a way it would remove any unused imports on save, but it can be removed manually.

File: io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java (hunk 6 above)
Comment (dwalluck) [reply io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java]: Yes, since we're writing out the settings file, we should want to keep the encoding of the original XML file, so I am trying to use `settings.getModelEncoding()` API to return the original encoding (which also should safely return `"UTF-8"` by default anyway as opposed to `null` anyway).

File: io/src/main/java/org/commonjava/maven/ext/io/CharsetAwareJacksonJsonProvider.java (hunk 11 above)
Comment (dwalluck) [reply io/src/main/java/org/commonjava/maven/ext/io/CharsetAwareJacksonJsonProvider.java]: I was able to make the jpath library load both UTF-16 and UTF-32 files successfully by overriding the `parse()` method. Before this, `DocumentContext.jsonString()` in `JSONIO.writeJSON()` would just throw an exception.

The point wasn't even to necessarily support these encodings as I don't know of a specific case using it, but to check that the round-trip was correct after we load and save JSON files.

Currently, the round-trip tests still throw some exception since the `FileIO.determineEOL()` method only works on single-byte encodings.

For example, `'\n'` is nromally encoded as `'\n' = 0x0A`, but it is two bytes, `0x00 0x0A`, in a UTF-16BE file. Therefore, the current method of looking back a single byte will not work.

But, I don't know if it's worth spending time to fix this now unless we ever come across such a file in the wild.

File: io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java (hunk 5 above)
Comment (rnc) [reply io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java]: I see what you are saying but 
```
    public final String toString() {
        return name();
    }
```
hence the change isn't really required?

File: io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java (hunk 6 above)
Comment (rnc) [reply io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java]: Ok cool

File: core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java (hunk 10 above)
Comment (rnc) [reply core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java]: I think I would prefer System.lineSeparator() if we are using that elsewhere (its then more consistent)

**Comment by rnc** [comment]
> > As a slight aside, is it your IDE reformatting and changing code that isn't directly related to your change? It makes the diffs much larger and harder to parse. Is there any way of only changing what is directly relevant?
> 
> Would it work on the pull request to click the icon at the top and select "Just for now
> Hide whitespace changes"?

Yeah but I think it would be better practice to minimise unnecessary changes?

File: io/src/main/java/org/commonjava/maven/ext/io/PomIO.java (hunk 13 above)
Comment (rnc) [reply io/src/main/java/org/commonjava/maven/ext/io/PomIO.java]: Huh. I'd completely fogotten that ManipEx had that .... lets discuss that offline.

File: core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java (hunk 10 above)
Comment (dwalluck) [reply core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java]: Maybe it makes sense to use `%n` only if there's an existing `String.format()` call or equivalent there and `System.lineSeparatror()` otherwise? The one time I'd say `%n` is easier to use is when it appears in the middle of lines that would otherwise have to all be manually concatenated.

**Comment by dwalluck** [comment]
> Yeah but I think it would be better practice to minimise unnecessary changes?

OK, but how do I do that now? Should I try to do `git diff --ignore-all-space upstream/master` and then re-apply the changes?

File: io/src/main/java/org/commonjava/maven/ext/io/CharsetAwareJacksonJsonProvider.java (hunk 11 above)
Comment (rnc) [reply io/src/main/java/org/commonjava/maven/ext/io/CharsetAwareJacksonJsonProvider.java]: I think its already a nice improvement. I would agree thats its not worth spending time resolving this when we haven't encountered a problem ever so far.

File: io/src/main/java/org/commonjava/maven/ext/io/FileIO.java (hunk 3 above)
Comment (rnc) [reply io/src/main/java/org/commonjava/maven/ext/io/FileIO.java]: Ok, lets remove it

File: core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java (hunk 10 above)
Comment (rnc) [reply core/src/test/java/org/commonjava/maven/ext/core/impl/VersionTest.java]: That makes some sense. String concatenation in tests shouldn't matter (and we also have the option of using the SLF4J MessageFormatter class).

File: io/src/main/java/org/commonjava/maven/ext/io/JSONIO.java [hunk 14]
```diff
@@ -15,50 +15,200 @@
...
+    }
+
+    public String detectEOL( File jsonFile ) throws ManipulationException
+    {
+        if ( charset == null )
+        {
+            detectEncoding( jsonFile );
+        }
+
+        if ( StandardCharsets.UTF_16BE.equals( charset ) )
```
dwalluck: This logic can all be simplified to `System.lineSeparator().getBytes( charset )`. The method is currently expected to return a string and not a byte array.

io/src/main/java/org/commonjava/maven/ext/io/XMLIO.java
dwalluck: I think it is just a coincidence that the code works. The argument that is supposed to be passed is the charset name and nothing would prevent `toString()` from returning some other information. The API says:

```every charset has a canonical name and may also have one or more aliases. The canonical name is returned by the name method of this class. Canonical names are, by convention, usually in upper case.```

rnc: Ok fine

rnc: Ok.

PR #710, opened 2019-11-14
Fix running of integration tests on Windows

dwalluck:


dwalluck:
Not all integration tests pass for me (with different failures on Linux and Windows), but this at least gets them running on Windows by fixing the Maven command.

I wonder if on Linux, too, it should prefer Maven home over trying to override the mvn in the `PATH`. This is probably more predictable across environments.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -248,9 +250,21 @@ static Integer runMaven( String commands, Map<String, String> params, String wor
         throws Exception
     {
         String stringParams = toJavaParams( params );
-        String commandMaven = String.format( "mvn %s %s ", commands, stringParams );
+        String commandMaven;
 
-        return runCommandAndWait(commandMaven, workingDir, MVN_LOCATION + "/bin");
+        if ( SystemUtils.IS_OS_WINDOWS )
+        {
+            commandMaven = String.format( "cmd.exe /C \"%s %s %s\"", MVN_LOCATION + File.separator + "bin"
```
dwalluck: Maybe it should repeat something like this below on line 262 and get rid of line 374? Overriding the `PATH` like this doesn't seem to work on Windows, as Windows doesn't exec in a shell it doesn't honor the `PATH`. I'd say it would work except that something else is being added to the `PATH`.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -64,7 +66,7 @@
 
     private static final String LOCAL_REPO = System.getProperty( "localRepositoryPath" );
 
-    static final File IT_LOCATION = new File( BUILD_DIR, "/it-cli");
+    static final File IT_LOCATION = new File( BUILD_DIR, "it-cli" );
```
dwalluck: Let the API join file components or use `File.separator`.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -248,9 +250,21 @@ static Integer runMaven( String commands, Map<String, String> params, String wor
...
 
-        return runCommandAndWait(commandMaven, workingDir, MVN_LOCATION + "/bin");
+        if ( SystemUtils.IS_OS_WINDOWS )
+        {
+            commandMaven = String.format( "cmd.exe /C \"%s %s %s\"", MVN_LOCATION + File.separator + "bin"
+                            + File.separator + "mvn.cmd", commands, stringParams );
+        }
+        else
+        {
+            commandMaven = String.format( "mvn %s %s", commands, stringParams );
```
dwalluck: Fixed extra space at end of command also.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -299,9 +313,15 @@ private static String toJavaParams( Map<String, String> params )
...
 
         StringBuilder stringParams = new StringBuilder();
-        for ( String key : params.keySet() )
+        Iterator<String> it = params.keySet().iterator();
+        while ( it.hasNext() )
         {
-            stringParams.append( String.format( "-D%s=%s ", key, params.get( key ) ) );
+            String key = it.next();
+            stringParams.append( String.format( "-D%s=%s", key, params.get( key ) ) );
+            if ( it.hasNext() )
```
dwalluck: Fix for extra space after last argument.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -342,36 +368,32 @@ private static Integer runCommandAndWait( String command, String workingDir, Str
...
+        File buildlog = new File(workingDir, "build.log");
 
-        BufferedReader stdout = new BufferedReader( new InputStreamReader( proc.getInputStream() ) );
-        BufferedReader stderr = new BufferedReader( new InputStreamReader( proc.getErrorStream() ) );
-        PrintWriter out = new PrintWriter( new BufferedWriter( new FileWriter( buildlog, true ) ) );
-
-        String line;
-        String errline = null;
-        while ( ( line = stdout.readLine() ) != null || ( errline = stderr.readLine() ) != null )
+        try ( BufferedReader stdout = new BufferedReader( new InputStreamReader( proc.getInputStream() ) );
```
dwalluck: Unrelated to the pull request, really, but fix possible resource leak.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -342,36 +368,32 @@ private static Integer runCommandAndWait( String command, String workingDir, Str
         String path = System.getenv( "PATH" );
         if (extraPath != null)
         {
-            path = extraPath + System.getProperty("path.separator") + path;
+            path = extraPath + File.pathSeparator + path;
         }
 
         Process proc = Runtime.getRuntime().exec(command, new String[] {"M2_HOME=" + MVN_LOCATION, "PATH=" + path}, new File(workingDir));
```
dwalluck: I am not sure why it doesn't just specify the full path to the `mvn` command it wishes to use rather than try to set the `PATH` and run `mvn` without any path.

I also think there could be an issue where the `mvn` script actually comes from the system path and `M2_HOME` refers to the local maven copy.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
rnc: @dwalluck Interestingly it used to be something like that (about 2 years ago) then I (!) changed it. Unfortunately I can't remember _why_ :-(  I suspect it might be historic - while older versions of mvn referenced M2_HOME I don't even see that referenced in modern versions ...... found it .... https://issues.apache.org/jira/browse/MNG-5607

So I would be happy to see this changed.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
rnc: See comment above

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
dwalluck: OK, I changed it to be the same as on Linux and use the path, so now it is `cmd.exe /X /C "mvn ..."`

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
rnc: @dwalluck Does this need to be changed as well for consistency?

dwalluck: There is also a reference to `M2_HOME` in `cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java`.

The scripts have a references to `MAVEN_HOME` and I see no reference to `M2_HOME`, so I don't think this is doing anything at all and we could try removing it.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -87,7 +87,7 @@
 public class Cli
 {
     private static final File DEFAULT_GLOBAL_SETTINGS_FILE =
-        new File( System.getenv( "M2_HOME" ), "conf" + File.separator + "settings.xml" );
+        new File( System.getProperty( "maven.home" ), "conf" + File.separator + "settings.xml" );
```
dwalluck: https://maven.apache.org/guides/introduction/introduction-to-profiles.html

Global
- Defined in the global Maven-settings (${maven.home}/conf/settings.xml).

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
```diff
@@ -342,36 +367,32 @@ private static Integer runCommandAndWait( String command, String workingDir, Str
         String path = System.getenv( "PATH" );
         if (extraPath != null)
         {
-            path = extraPath + System.getProperty("path.separator") + path;
+            path = extraPath + File.pathSeparator + path;
         }
 
-        Process proc = Runtime.getRuntime().exec(command, new String[] {"M2_HOME=" + MVN_LOCATION, "PATH=" + path}, new File(workingDir));
```
dwalluck: Or set `MAVEN_HOME`?

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
rnc: Hmmm. On second thoughts, given I occasionally run on Maven 3.0.5 to check compatibility I think its better to leave it. ( I think it was removed >= 3.4 (?) )

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Looking at https://maven.apache.org/shared/maven-invoker/usage.html while in general I think this should be ok I am beginning to think especially if we want to test on older versions occasionally we might need some backwards compatibility here. Perhaps belt/braces approach and set all of them?!!

rnc: Similar to how https://maven.apache.org/shared/maven-invoker/usage.html does ?

rnc: Lets discuss today on IRC

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/TestUtils.java
dwalluck: I checked and we use version 3.5.0 for the ITs, and this appears to be the first version where `M2_HOME` is gone from the scripts. For in the Java code, the property `${maven.home}` would be the right option for all versions since all it does is set `-Dmaven.home=$M2_HOME`.

The code we use for the settings isn't the same as upstream which also allows overriding the location, see https://github.com/apache/maven/blob/8d221f97fc72fbd9099e082600590fdf6b85143c/maven-core/src/main/java/org/apache/maven/settings/DefaultMavenSettingsBuilder.java#L68.

The latest versions have a new property, `maven.conf`. Old versions used `${maven.home}/conf`. All allow override via `MavenSettingsBuilder.ALT_GLOBAL_SETTINGS_XML_LOCATION = org.apache.maven.global-settings`.

rnc: Ok nice investigation! Lets purge M2_HOME then.

dwalluck: So when you said 3.0.5, did you mean 3.5.0? As you said, it will depend on the minimum Maven version requirement.

I still think we won't remove `M2_HOME`, but replace with `MAVEN_HOME` as the scripts do `-Dmaven.home=${MAVEN_HOME}`.

rnc: But above the line `-Dmaven.home...` they do `MAVEN_HOME=`dirname "$PRG"`/..` ?

PR #712, opened 2019-11-21
PRODTASKS-737 Simplify and make use of SLF4J style {} in custom Excep…

rnc:
…tions


common/src/main/java/org/commonjava/maven/ext/common/util/ProjectComparator.java
```diff
@@ -430,7 +430,7 @@ private static void append (StringBuilder builder, String message, Object ...arg
                 }
                 default:
                 {
-                    throw new ManipulationException( "Invalid type " + type.toString() );
+                    throw new ManipulationException( "Invalid type {}", type.toString() );
```
dwalluck: There are several references to `toString()` in the new code as well as the original. I think we just want to pass the object and defer the call to `toString()`.

rnc: Good spot! Completely agree these should be removed.

PR #713, opened 2019-11-23
Fix readModelsForManipulation call in PomIO.parseProject

dwalluck:
The method readModelsForManipulation was passing
pom.getAbsoluteFile which could lead to failures later on in
cases where pom.getAbsoluteFile was not equal to
pom.getCanonicalFile.

### DIFF
--- a/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
+++ b/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
@@ -68,10 +68,17 @@ public class PomIO
     private static final Logger logger = LoggerFactory.getLogger( PomIO.class );
 
 
-    public List<Project> parseProject (final File pom) throws ManipulationException
+    public List<Project> parseProject( final File pom ) throws ManipulationException
     {
-        final List<PomPeek> peeked = peekAtPomHierarchy(pom);
-        return readModelsForManipulation( pom.getAbsoluteFile(), peeked );
+        final List<PomPeek> peeked = peekAtPomHierarchy( pom );
+        try
+        {
+            return readModelsForManipulation( pom.getCanonicalFile(), peeked );
+        }
+        catch ( IOException e )
+        {
+            throw new ManipulationException( "Error getting canonical file", e );
+        }
     }
 
     /**


rnc:
Do you have any examples of this causing an issue?

dwalluck:
> Do you have any examples of this causing an issue?

I believe that `getCanonicalPath()` does not equal `getAbsolutePath()` whenever we have a symlink in the path containing our `pom.xml`. If I were to set up a test case, I think I would try that.

The failing test was

```
[ERROR] shouldRemoveProjectInGroovyScript(org.commonjava.maven.ext.core.impl.InitialGroovyManipulatorTest)  Time elapsed: 1.873 s  <<< ERROR!
org.commonjava.maven.ext.common.ManipulationException: First project is not execution root
```

except that it was the execution root and `File.equals()` was failing because it was comparing two different things.

It happened when I set up Travis CI under Mac OS X in my local fork. If you want to run the build Mac OS X here, I can set up another pull request with my updated `.travis.yml`.

rnc:
What are you doing up?! There are other places using absolute that have some arguments for using canonical. I imagine a test case could be constructed for this. Why are you setting up a MacOS build?

dwalluck:
The believe the exact bug is that on https://github.com/release-engineering/pom-manipulation-ext/blob/8169e4ce035ecece0d36867395b3d41afbdc26b8/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java#L74 it is passing `getAbsoluteFile()`, but earlier on https://github.com/release-engineering/pom-manipulation-ext/blob/8169e4ce035ecece0d36867395b3d41afbdc26b8/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java#L300 it stored `getCanonicalFile()` in the list.

dwalluck:
> Why are you setting up a MacOS build?

My original idea was to set up a Windows build so that I could automate running the integration tests to test the URLs. I guess Travis CI doesn't support Windows, but a couple additional lines got me a Mac OS X build for free so as to not be a total waste of effort. Windows requires setting up AppVeyor for CI (`appveyor.yml`).

rnc:
Ok. I've tried creating a test and validated it - please see https://github.com/release-engineering/pom-manipulation-ext/compare/master...rnc:713?expand=1 which is layered on top of this PR.

dwalluck:
So `testVerifyCanonicalExecutionRoot()` actually fails without the fix? That's good.

I think that you can check out my personal repository and commit directly on this pull request if you want instead of creating your own branch. Then it would show up here.

I was going to make a pull request for the URLs separately. It's almost exactly like yours with one additional change in `BaseGroovyManipulator` and I had to change the `groovyScripts=` URLs to avoid an exception.

I can open a PR for that and you can decide if you want to commit on top of it or we can rebase it once this is in.

PR #714, opened 2019-11-23
Fix file URLs

dwalluck:


cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -448,10 +449,10 @@ private void createSession( File target, File settings )
             {
                 // No, this is not a typo. If current default is null, supply new local and global.
                 // This function passes in settings to make it easier to test.
-                this.settings = settings = new File( System.getProperty( "user.home" ), ".m2/settings.xml" );
+                this.settings = settings = new File( System.getProperty( "user.home" ), ".m2" + File.separator + "settings.xml" );
```
dwalluck: This is obviously a fix for file separator not URL.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -465,7 +466,7 @@ private void createSession( File target, File settings )
 
             if ( ar != null)
             {
-                ar.setUrl( "file://" + req.getLocalRepositoryPath() );
+                ar.setUrl( req.getLocalRepositoryPath().toPath().toUri().toString() );
```
dwalluck: I assume that `File.toURI()` is the same as `Path.toUri()`, but just to be safe I converted it to a `Path` first. There is something int he javadoc about `File.toURL()` not being safe, in any case.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -94,10 +94,12 @@
                 for ( final String script : scripts )
                 {
                     File found;
-                    if ( script.startsWith( "http" ) || script.startsWith( "file" ))
+                    logger.info( "Processing groovy script {}", script );
+                    if ( script.startsWith( "http:" ) || script.startsWith( "file:" ) )
```
dwalluck: Missing `:` here to mark the protocol.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -94,10 +94,12 @@
...
                     {
-                        logger.info( "Attempting to read URL {} ", script );
-                        found = fileIO.resolveURL( new URL( script ) );
+                        URL url = new URL( script );
+                        logger.info( "Attempting to read URL {}", url );
+                        found = fileIO.resolveURL( url );
```
dwalluck: This is the line that does not like the `groovyScripts=file://` URLs which gives `UnknownHostException: C` on Windows later when it does `url.openStream()` in `fileIO.resolveURL()`.

rnc: I don't see why this is needed?

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
rnc: I don't see why this is needed?

integration-test/src/it/circular-dependencies-test-second/test.properties
```diff
@@ -17,5 +17,5 @@
 restURL=http://127.0.0.1:8089
 versionIncrementalSuffix=redhat
 dependencySource=rest
-groovyScripts=file://${basedir}/src/it/circular-dependencies-test-second/Sample.groovy
```
rnc: The `file://` as I thought we discussed previously is an acceptable standard so I don't see why this needs changing?

integration-test/src/it/groovy-manipulator-file/verify.groovy
```diff
@@ -30,7 +30,7 @@ assert !jsonString.contains ("resolved:")
 def buildLog = new File( basedir, 'build.log' )
 def message = 0
 buildLog.eachLine {
-   if (it.contains( "groovyScripts file:///")) {
+   if (it.contains( "groovyScripts file:")) {
```
rnc: Why this change?

integration-test/src/it/groovy-manipulator-file/test.properties
```diff
@@ -16,4 +16,4 @@
 
 
 # The colon needs to be escaped to prevent the invoker plugin from misinterpreting it
-groovyScripts=file://${basedir}/src/it/setup/depMgmt1/Sample.groovy
```
rnc: As per above

integration-test/src/it/remote-dependency-management-embedded-props-strict/test.properties
```diff
@@ -21,4 +21,4 @@ dependencyManagement=org.commonjava.maven.ext.integration-test\:depMgmt2\:1.0.0-
 version.suffix=redhat-1
 overrideTransitive=false
 strictAlignment=true
-groovyScripts=file://${basedir}/src/it/remote-dependency-management-embedded-props-strict/inlineProp.groovy
```
rnc: As per above

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
dwalluck: If you want to check if it's a URL, then the URL must start with `scheme + ':'`, or otherwise strings such as `"file.txt"` will match, which is unintended, and it could even match before the the final `else` case.

It also matches what it used elsewhere in the code:

https://github.com/release-engineering/pom-manipulation-ext/blob/8606c8b023b8a60333e41495d3ad38e466254491/core/src/main/java/org/commonjava/maven/ext/core/impl/RepoAndReportingRemovalManipulator.java#L264

https://github.com/release-engineering/pom-manipulation-ext/blob/8606c8b023b8a60333e41495d3ad38e466254491/io/src/main/java/org/commonjava/maven/ext/io/resolver/GalleyInfrastructure.java#L238

https://github.com/release-engineering/pom-manipulation-ext/blob/8606c8b023b8a60333e41495d3ad38e466254491/io/src/main/java/org/commonjava/maven/ext/io/resolver/GalleyInfrastructure.java#L266

https://github.com/release-engineering/pom-manipulation-ext/blob/8606c8b023b8a60333e41495d3ad38e466254491/io/src/main/java/org/commonjava/maven/ext/io/resolver/MavenLocationExpander.java#L115

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
dwalluck: I was trying to see if there was a difference between the given input URL which is `script` and the url after it has been parsed which is `url`. This was for debugging, so `logger.debug` could have been used.

integration-test/src/it/circular-dependencies-test-second/test.properties
dwalluck: URLs beginning with `file://` are definitely incorrect, but I can revert the changes to the properties files and try to fix up the invalid URL in the code instead. I can add a test case to show the failure.

Note that such a URL parses, so in that sense it is _valid_, but it's _semantically_ incorrect in that it expects an `authority` (hostname) to come after two slashes, which is true in the case of a network URL, but a file URL should have no `authority`

integration-test/src/it/groovy-manipulator-file/verify.groovy
dwalluck: Because if `${basedir}` does not begin with a `/` then `file://${basedir}` won't have three slashes and this check will fail, so it should be sufficient to check for `file:` which is what seems to be used everywhere else in the code anyway.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
rnc: Ok

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: I've changed this code so I think it might be good to rebase to master please

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
rnc: Ok sure, I guess being consistent is sensible.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
rnc: As this can convert the original `script` value I think it would be better kept as it was.

integration-test/src/it/groovy-manipulator-file/verify.groovy
rnc: Ok

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: As per above, I think this needs rebasing.

integration-test/src/it/circular-dependencies-test-second/test.properties
rnc: Lets discuss offline.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
dwalluck: There's no code that changed here. I just moved it to two separate lines so I could log the difference between the values. It looks like `URL.toString()` at the very least was printing only one slash in some cases, which is strange. In any case, I can revert it.

integration-test/src/it/circular-dependencies-test-second/test.properties
dwalluck: All I know is that `file:` with no slashes appeared to work on all OSes via the CI. With two or three slashes, there will need to be code to fix up the URI on Windows.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -69,6 +71,17 @@
 
     public abstract int getExecutionIndex();
 
+    String fixFileURL( final String script )
```
rnc: This means this method will be exposed to even the Gradle Manipulator tooling which is not ideal - given its internal only. Can we discuss offline please.

dwalluck:
Obsoleted by #717

PR #718, opened 2019-11-27
PRODTASKS-701 Code Coverage

dwalluck:


dwalluck:
Does this need `'[ "$TRAVIS_OS_NAME" == "linux" ] && bash'` to prevent multiple uploads? I am not sure if codecov just takes the first one that finishes.

dwalluck:
I don't think we'll see anything until the codecov support is merged into master. It looks like it successfully uploaded in the Travis logs.

pom.xml
```diff
@@ -77,7 +77,7 @@
     <javaVersion>1.8</javaVersion>
     <javaTestVersion>${javaVersion}</javaTestVersion>
 
-    <mavenVersion>3.5.0</mavenVersion>
```
rnc: Why this change?

pom.xml
```diff
@@ -95,44 +95,55 @@
       <dependency>
         <groupId>org.commonjava.maven.ext</groupId>
         <artifactId>pom-manipulation-common</artifactId>
-        <version>3.9-SNAPSHOT</version>
+        <version>${project.version}</version>
       </dependency>
       <dependency>
```
rnc: Why this change?

pom.xml
dwalluck: No problem. I didn't mean to leave this in. I was just checking if the tests passed regardless of the version used.

pom.xml
dwalluck: This was to make sure that these dependencies automatically keep the version in sync with the project version since these dependencies should always come from the project build itself and not an external maven repository.

dwalluck:
> * What does this cover (i.e. coverage scope)?

I know that when the integration tests module is included, the missed instructions goes from about 12,000 to about 4,000.

pom.xml
rnc: The release plugin handles that - see http://central.maven.org/maven2/org/commonjava/maven/ext/pom-manipulation-parent/3.8.1/pom-manipulation-parent-3.8.1.pom for example. Also using $project.version is really not advised (which is why PME inlines it when it encounters it). Please revert this.

pom.xml
rnc: Ok

rnc:
> > * What does this cover (i.e. coverage scope)?
> 
> I know that when the integration tests module is included, the number of missed instructions goes from about 12,000 to about 4,000.

I think you're misunderstanding me. As your PR stands - is this just covering unit tests or unit tests and integration tests? I would have expected to see e.g. jacoco goals of `prepare-agent-integration`?

rnc:
> Does this need `'[ "$TRAVIS_OS_NAME" == "linux" ] && bash'` to prevent multiple uploads? I am not sure if codecov just takes the first one that finishes.

Can you elaborate please?

pom.xml
dwalluck: > The release plugin handles that - see http://central.maven.org/maven2/org/commonjava/maven/ext/pom-manipulation-parent/3.8.1/pom-manipulation-parent-3.8.1.pom for example. Also using $project.version is really not advised (which is why PME inlines it when it encounters it). Please revert this.

This doesn't make a lot of sense to me as there are already five or more version properties defined in the pom, presumably for the same reason, to avoid errors and repeatedly hardcoding the same value over and over.

dwalluck:
> Can you elaborate please?

If you look at the Travis CI logs, there's an upload of the coverage report for every build that finishes instead of one per pull request. So which one is it using? Just randomly the first build to finish? So, I was thinking how this could be limited to one report.

pom.xml
rnc: Those aren't using project.version and, as I said, the release plugin handles transforming this. Its done on purpose ;-)

rnc:
> > > * What does this cover (i.e. coverage scope)?
> > 
> > 
> > I know that when the integration tests module is included, the number of missed instructions goes from about 12,000 to about 4,000.
> 
> I think you're misunderstanding me. As your PR stands - is this just covering unit tests or unit tests and integration tests? I would have expected to see e.g. jacoco goals of `prepare-agent-integration`?

I think you missed this comment :-)

PR #722, opened 2019-12-11
NCL-5068 Allow passing custom headers to REST endpoints

dwalluck:


core/src/main/java/org/commonjava/maven/ext/core/ManipulationSession.java
```diff
@@ -140,6 +145,16 @@ public void setProjects( final List<Project> projects )
...
+    public void setHeaders( final Map<String, String> headers )
+    {
+        this.headers = headers;
+    }
+
+    public Map<String, String> getHeaders()
+    {
+        return headers == null ? Collections.emptyMap() : Collections.unmodifiableMap( headers );
+    }
+
```
rnc: Not required as per conversation on IRC

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -541,4 +558,35 @@ private Settings parseSettings( File settings ) throws ComponentLookupException,
...
+        public String getName()
+        {
+            return name;
+        }
+
+        public String getValue()
+        {
+            return value;
+        }
+    }
```
rnc: Not required as per conversation on IRC

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -54,7 +54,7 @@ public void initialise( Properties userProps )
                                                                    String.valueOf( DefaultTranslator.CHUNK_SPLIT_COUNT ) ) );
         restSuffixAlign = Boolean.parseBoolean( userProps.getProperty( "restSuffixAlign", "true" ) );
 
-        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, vState.getIncrementalSerialSuffix() );
+        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, vState.getIncrementalSerialSuffix(), session.getHeaders() );
```
rnc: I think we can retrieve the values from user properties

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -108,14 +110,29 @@
...
+     * @param endpointUrl is the URL to talk to.
+     * @param restMaxSize initial (maximum) size of the rest call; if zero will send everything.
+     * @param restMinSize minimum size for the call
+     * @param repositoryGroup the group to pass to the endpoint.
+     * @param incrementalSerialSuffix the suffix to pass to the endpoint.
+     */
+    public DefaultTranslator( String endpointUrl, int restMaxSize, int restMinSize, String repositoryGroup, String incrementalSerialSuffix )
+    {
+        this ( endpointUrl, restMaxSize, restMinSize, repositoryGroup, incrementalSerialSuffix,
+               Collections.emptyMap() );
```
rnc: Why do we need another ctor?

dwalluck:
> # [Codecov](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=h1) Report
> > Merging [#722](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=desc) into [master](https://codecov.io/gh/release-engineering/pom-manipulation-ext/commit/5020a177d727f18348acc9933d0d708f8c3171e7?src=pr&el=desc) will **increase** coverage by `30.69%`.
> > The diff coverage is `44.11%`.

I am not sure why it is doing this. It looks like with ITs vs. without.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
dwalluck: > Why do we need another ctor?

It was used to pass the headers through to this class, but shouldn't be needed if using properties.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
dwalluck: > Not required as per conversation on IRC

This is used to split the headers into their components. Unless I am missing something, if we use a single property, then we still need to split the many headers into single headers, and then split the header into name-value pairs since Unirest expects a `java.util.Map` for multiple headers.

rnc: Ok yes I would agree .... but the CLI shouldn't be doing it as it will already store the property - which I think I gave a potential suggestion for - more likely it would be inside RESTState.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Ok!

rnc:
> > # [Codecov](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=h1) Report
> > > Merging [#722](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=desc) into [master](https://codecov.io/gh/release-engineering/pom-manipulation-ext/commit/5020a177d727f18348acc9933d0d708f8c3171e7?src=pr&el=desc) will **increase** coverage by `30.69%`.
> > > The diff coverage is `44.11%`.
> 
> I am not sure why it is doing this. It looks like with ITs vs. without.

Agreed. Annoying :-( Looks like something isn't quite right...

rnc:
> > > # [Codecov](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=h1) Report
> > > > Merging [#722](https://codecov.io/gh/release-engineering/pom-manipulation-ext/pull/722?src=pr&el=desc) into [master](https://codecov.io/gh/release-engineering/pom-manipulation-ext/commit/5020a177d727f18348acc9933d0d708f8c3171e7?src=pr&el=desc) will **increase** coverage by `30.69%`.
> > > > The diff coverage is `44.11%`.
> > 
> > 
> > I am not sure why it is doing this. It looks like with ITs vs. without.
> 
> Agreed. Annoying :-( Looks like something isn't quite right...

Ahh - I bet its because the master branch is the 'default', and that is built automatically on PR merge but **without** ITs - hence the coverage changes.

rnc:
@dwalluck  I've pushed a change ; but not sure how to get the code-cov to regenerate?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -191,14 +208,15 @@ private void autoPartition(List<ProjectVersionRef> projects, Queue<Task> queue)
...
 
-        logger.trace( "Called findBlacklisted to {} with {}", blacklistEndpointUrl, ga );
+        logger.debug( "Called findBlacklisted to {} with {} and custom headers {}", blacklistEndpointUrl, ga, headers );
 
         try
         {
             r = Unirest.get( blacklistEndpointUrl )
                        .header( "accept", "application/json" )
                        .header( "Content-Type", "application/json" )
                        .header( "Log-Context", getHeaderContext() )
```
rnc: I think we should remove getHeaderContext and merge it in with `headers` ?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
dwalluck: > Ok!

I moved the code to `RESTState`, but it still needs to pass the `restHeaders` like all these other arguments here are (which also come from user properties). Is the issue that it's adding an argument, or just that there should be a single constructor and not two? Most of the calls wouldn't pass any headers, just the call from the `RESTState` does.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
dwalluck: > I think we should remove getHeaderContext and merge it in with `headers` ?

Removing the log context code currently causes a lot of tests to have assertion failures and I haven't figured it out yet.

io/src/test/java/org/commonjava/maven/ext/io/rest/BlacklistTranslatorTest.java
```diff
@@ -52,8 +54,11 @@ public void before()
     {
         LoggerFactory.getLogger( BlacklistTranslatorTest.class ).info( "Executing test " + testName.getMethodName() );
 
+        Map<String, String> headers = new LinkedHashMap<>( 1 );
+        headers.put( "Foo", "bar" );
```
dwalluck: I am trying to add a fake header to test that it's passed through. It can be seen in the logs (of the base class), but I don't think there's any way to get at the headers and assert something here.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Can you send me which tests? Junit or integration?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -54,7 +62,16 @@ public void initialise( Properties userProps )
                                                                    String.valueOf( DefaultTranslator.CHUNK_SPLIT_COUNT ) ) );
         restSuffixAlign = Boolean.parseBoolean( userProps.getProperty( "restSuffixAlign", "true" ) );
 
-        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, vState.getIncrementalSerialSuffix() );
+        String restHeadersProperty = userProps.getProperty( "restHeaders", "" );
+        if ( !StringUtils.isEmpty( restHeadersProperty ) )
+        {
+            restHeaders = Arrays.stream( restHeadersProperty.split( "," ) )
+                                .map( Header::new )
+                                .collect( Collectors.toMap( Header::getName, Header::getValue,
```
rnc: Isn't using Header a bit 'overkill' - why not split on comma and split on colon ?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: I would be curious to see what tests fail with this removed.

io/src/test/java/org/commonjava/maven/ext/io/rest/BlacklistTranslatorTest.java
rnc: `HttpHeaderHeaderTest` hacks around it. Also `RESTParametersTest` creates a dummy `AbstractHandler` which could also be used (as the headers can be extracted from `HttpServletRequest`)

dwalluck: OK, I created a new test based on `RESTParametersTest`.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
dwalluck: It may be better to make a separate pull request for removing the log context option.

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
dwalluck: I agree. I thought it might be easier to handle the edge cases like a null or empty header name, but I've converted it to not use a class.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Ok, I guess I'm ambivalent about that. As long as the log-context option is removed (and in the documentation) - of course we will still 'implicitly' support it via your new restHeaders option.

rnc:
Looks good ; I guess this is ready for review now?

rnc:
One other thing I think would be good ( although can be a subsequent PR) is unit tests for the RESTstate changes to parse the header collection.

dwalluck:
> One other thing I think would be good ( although can be a subsequent PR) is unit tests for the RESTstate changes to parse the header collection.

The test should start by parsing the property value and not the just using a preexisting map like it is now. I will see if I can update it.

rnc:
> > One other thing I think would be good ( although can be a subsequent PR) is unit tests for the RESTstate changes to parse the header collection.
> 
> The test should start by parsing the property value and not the just using a preexisting map like it is now. I will see if I can update it.

Given that it would require a separate test (with a dummy'd up `RESTState`) my vote would be move this to non-draft and make a subsequent PR for the extra test - it could even be an integration test. I have added a layered PR for the removal of the log context code.

PR #727, opened 2020-01-07
NCL-5068 Document REST headers feature

dwalluck:


rnc:
Change looks fine apart from if we merge it now people will think its immediately available. Therefore we either need to:

1. Use a warning note regarding versioning such as is used https://release-engineering.github.io/pom-manipulation-ext/guide/groovy.html#inlineproperty (i.e. available after 3.8.1).

1. Consider switching to a current stable versus development branch
   - This is rather more complex as it would require restructuring e.g.
     - https://github.com/koumoul-dev/gh-pages-multi
     - https://stackoverflow.com/questions/32629223/github-pages-dev-branch
   - We would also have to think about when dev became stable, how to label features that only appeared in version X

Given the last point, I think option (1) is the best choice. Thoughts?

dwalluck:
> Given the last point, I think option (1) is the best choice. Thoughts?

Actually, I had originally thought of doing it as a note, so maybe we should remove the section and just add a second note there to the the existing REST section.

rnc:
> > Given the last point, I think option (1) is the best choice. Thoughts?
> 
> Actually, I had originally thought of doing it as a note, so maybe we should remove the section and just add a second note there to the the existing REST section. I guess the question is how much of it should be a note or just add a note about the version to the existing text?

Possibly - not sure which section you are talking about tbh.

PR #730, opened 2020-01-10
NCL5368 Add property annotation system

rnc:


rnc:
@dwalluck  I think this is ready for review now

annotation/src/main/java/org/commonjava/maven/ext/annotation/ConfigValueProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+
+    void generateCode() throws IOException
+    {
+        ClassName hashMap = ClassName.get( "java.util", "HashMap");
+        ClassName str = ClassName.get( String.class );
+        ParameterizedTypeName mainType = ParameterizedTypeName.get( hashMap, str, str );
+
+        Set<CodeBlock> contents = new HashSet<>(  );
+        varResults.forEach( ( key, value ) -> contents.add(
+                        CodeBlock.builder().addStatement( "allConfigValues.put($S, $S);", key, value ).build() ) );
```
dwalluck: The `;` shouldn't be necessary here as the statement already apparently includes one, leading to double `;;`.

annotation/src/main/java/org/commonjava/maven/ext/annotation/ConfigValueProcessor.java
```diff
@@ -0,0 +1,139 @@
...
+
+        TypeSpec ConfigList = TypeSpec.classBuilder( "ConfigList" )
+                                      .addModifiers( Modifier.PUBLIC, Modifier.FINAL )
+                                      .addField( FieldSpec.builder
+                                                      ( mainType, "allConfigValues", Modifier.STATIC, Modifier.FINAL, Modifier.PUBLIC )
+                                                          .initializer( "new HashMap<>()" )
```
dwalluck: A small optimization can be to set the size of to `allConfigValues` to `contents.size()`.

PR #737, opened 2020-02-09
NCL-5468 Add pluginOverride. Refactor code.

rnc:


rnc:
@dwalluck  Hi. Was looking at the code to see if its possible to add `pluginOverride` (like `dependencyOverride` ). Have a look - its still WIP (no tests yet) but it seems to work.

common/src/main/java/org/commonjava/maven/ext/common/model/ArtifactPluginWrapper.java
```diff
@@ -0,0 +1,42 @@
...
+ * Simple wrapper to allow a Plugin to be treated as an ArtifactRef which allows for more generic code.
+ */
+public class ArtifactPluginWrapper extends SimpleArtifactRef
+{
+    @Getter
+    private final Plugin original;
+
+    public ArtifactPluginWrapper( Plugin p )
+    {
+        super (p.getGroupId(), p.getArtifactId(), p.getVersion(), null, null);
```
dwalluck: Should type be set to `"maven-plugin"`?

core/src/main/java/org/commonjava/maven/ext/core/groovy/ASTTransformer.java
```diff
@@ -194,18 +206,11 @@ private void changeBaseScriptType(final AnnotatedNode parent, final ClassNode cN
 
         List<AnnotationNode> annotations = parent.getAnnotations( DEPRECATED_COMMAND_TYPE );
         if (cNode.getAnnotations( DEPRECATED_COMMAND_TYPE ).isEmpty()) { // #388 prevent "Duplicate annotation for class" AnnotationFormatError
-
-            // TODO : handle scenario where the annotation is not fully qualified which comes through as a variable
```
dwalluck: Anything added for these?

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/invoker/DefaultExecutionParser.java
```diff
@@ -40,12 +40,7 @@
         add( ExecutionParser.SYSTEM_PROPERTIES_HANDLER );
     }};
 
-    private List<ExecutionParserHandler> handlers;
-
-    public DefaultExecutionParser()
-    {
-        this( new ArrayList<ExecutionParserHandler>() );
-    }
+    private final List<ExecutionParserHandler> handlers;
```
dwalluck: Would it make sense to use `Collections.unmodifiableList()` here, too?

core/src/main/java/org/commonjava/maven/ext/core/groovy/ASTTransformer.java
rnc: No, this code came from the original classes, so its just cleaning it up

common/src/main/java/org/commonjava/maven/ext/common/model/ArtifactPluginWrapper.java
rnc: No I don't think so. The Atlas type is not referring to the pom packaging type but the 'type' of the artifact e.g https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-compiler-plugin/3.8.1/ is a jar.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/invoker/DefaultExecutionParser.java
rnc: I've taken another look and simplified the code more. It now implicitly uses a fixed size list.

PR #744, opened 2020-02-24
NCL5481 - dependencyOverride and deprecated properties

rnc:


core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -118,6 +120,34 @@ public void init( final ManipulationSession session )
...
+                                forEach(up ->
+                                {
+                                    logger.warn
+                                            ("Removing key {} (as deprecated) from user properties (with matcher of {})",
+                                                    p,
+                                                    up.getKey());
+                                    session.getUserProperties().remove(p);
+                                });
+                    }
+                });
```
dwalluck: A suggestion here is to print (all) warnings and then fail. At this point, user has two options to fix it, enable the deprecated options, or fix that particular option that has the warning. If we had access to a map that had equivalent non-deprecated options, we could suggest that option.

rnc: It now throws an exception below.

rnc:
Still need to do documentation!

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -82,11 +84,15 @@
 
     public static final String REPORT_JSON_OUTPUT_FILE = "reportJSONOutputFile";
 
+    // TODO: Config index value
```
rnc: Chicken/egg ; can't fill this in till we do the docs and they are merged and then the index exists.

PR #751, opened 2020-03-04
PRODTASKS-694: Apply relocations to plugins

dwalluck:


core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -128,13 +161,66 @@ private boolean apply( final Project project, final Model model ) throws Manipul
...
         }
+
+        result |= updateAllPlugins( model, project, relocations );
+
+        return result;
+    }
+
+    private boolean updateAllPlugins( final Model model, final Project project, final WildcardMap<ProjectVersionRef> relocations ) throws ManipulationException
+    {
+        Map<String, Plugin> pluginMap = getManagedPluginMap( model );
```
rnc: Why can't we use https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/model/Project.java#L381 ?

dwalluck: I'll change it to use that and remove the older code from the file. I believe that code is not handling profiles anyway.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -128,13 +161,66 @@ private boolean apply( final Project project, final Model model ) throws Manipul
...
+        List<PluginReference> refs =  findPluginReferences( model, project, pluginMap );
+        boolean result = updatePlugins( relocations,  refs );
+
+        pluginMap = getPluginMap( model );
```
rnc: As per above use the Project API - it also resolves any variables for you.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
...
+    }
+
     @Override
     public int getExecutionIndex()
     {
         return 15;
     }
+
+    private final class PluginReference
+    {
```
rnc: Question : are we needing to store the Node objects in order to be able to update them transparently ?

I think a couple of sentences of javadoc on this class would be useful :-)

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
...
+
+            return result;
+        }
+
+        return Collections.emptyMap();
+    }
+
+    /**
+     * Go through the plugin / plugin-execution configurations and find references to the <code>skip</code> parameter for the given Maven plugin
+     * instance.
```
rnc: I think this doc isn't right ?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
...
+     * @return the configurations
+     */
+    private Map<ConfigurationContainer, String> findConfigurations( final Plugin plugin )
```
rnc: Any particular reason to key off ConfigurationContainer rather than Plugin?

rnc:
As a more general comment I think we'll have to consider factoring some of this out so it can be used by the PluginManipulator (especially with the upcoming work) but that can be done in future work.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
...
+
+            this.versionNode = versionNode;
+
+            if ( versionNode != null )
+            {
+                setVersion( versionNode.getTextContent() );
+            }
+        }
+
+        public ConfigurationContainer getContainer()
```
rnc: Don't really need getter/setter for inner class I think?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -128,13 +161,66 @@ private boolean apply( final Project project, final Model model ) throws Manipul
...
+
+        logger.debug( "Got {} relocations for plugins", size );
+
+        boolean result = false;
+
+        for ( int i = 0; i < size; i++ )
+        {
+            final Iterator<PluginReference> it = dependencies.iterator();
+
+            while ( it.hasNext() )
```
rnc: Could use for loop to stop intellij complaining :-)

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
         }
     }
 
+    private PluginReference updatePluginGAVs( final Map.Entry<ConfigurationContainer, String> entry, final NodeList children )
```
rnc: Naming : it seems we're not updating as such ; rather locating any plugins?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -203,9 +289,298 @@ private void updateDependencyExclusion( ProjectVersionRef depPvr, ProjectVersion
...
+
+            logger.debug( "Adding plugin reference: {}", ref );
+
+            return ref;
+        }
+
+        return null;
+    }
+
+    public List<PluginReference> findPluginReferences( final ModelBase base, final Project project, Map<String, Plugin> pluginMap ) throws ManipulationException
```
rnc: `base` is never used?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -128,13 +161,66 @@ private boolean apply( final Project project, final Model model ) throws Manipul
...
+
+        pluginMap = getPluginMap( model );
+        refs = findPluginReferences( model, project, pluginMap );
+
+        result |= updatePlugins( relocations, refs );
+
+        return result;
+    }
+
+    private boolean updatePlugins( final WildcardMap<ProjectVersionRef> relocations, final List<PluginReference> dependencies ) throws ManipulationException
```
rnc: A few questions:

- What happens if the relocation includes a version? 
- What happens if the plugin includes a version ?
- What happens if the version is a variable?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
rnc: Good point ; there are specialised functions in the Project class to return fully resolved set from profiles as well.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: It might be possible to update inline.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: The code for this came from another class. I am not even sure the map key is being used, just the value. But it wants to get the XML for the plugin's config section in order to parse and update it.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: I am not getting an error/warning here.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: It is finding a GAV tag set inside a plugin's configuration section.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: OK.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: There is definitely some issue here. I can't find what the dependency override is doing with the versions to begin with.

The code created a key from the found GA (no V) and looks up the GA in `WildcardMap<ProjectVersionRef> relocations`.

I think there must also be an issue with wildcard for `artifactId` as well.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: OK, it is agreed to make this change.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: OK, it is agreed to make this change.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: OK, it is agreed to make this change.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: Right. This was taken from other code, so maybe there was some reason for keying the map like this. I am not finding a reference to the key, in which case, we may be able to change `Map<ConfigurationContainer, String>` to `List<String>` as well.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: It's using some of these methods on lines 206--210.

dwalluck: Getters and setters have been removed.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
dwalluck: Changed name to `findPluginReference`.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -177,6 +221,53 @@ private boolean updateDependencies( WildcardMap<ProjectVersionRef> relocations,
...
+
+                    if ( !relocation.getArtifactId().equals( WildcardMap.WILDCARD ) )
+                    {
+                        ref.artifactIdNode.setTextContent( relocation.getArtifactId() );
+                    }
+
+                    if ( ref.versionNode != null && !relocation.getVersionString().equals( WildcardMap.WILDCARD ) )
+                    {
+                        ref.versionNode.setTextContent(  relocation.getVersionString() );
+                    }
```
dwalluck: This code is supposed to handle wildcards and sections that contain a version.

integration-test/src/it/relocation-plugin/verify.groovy
```diff
@@ -0,0 +1,42 @@
...
+def dependency = pom.dependencies.dependency.find { it.groupId.text() == "junit" }
+assert dependency != null
+assert dependency.artifactId.text() == "junit"
+assert dependency.version.text() == "4.1"
+
+dependency = pom.dependencies.dependency.find { it.groupId.text() == "org.slf4j" }
+assert dependency != null
+assert dependency.artifactId.text() == "slf4j-api"
+assert dependency.version.text() == "1.7.30"
+
```
rnc: The checks above do not accurately check that the relocations for junit/slf4j worked - they verify the dependencies but _not_ the artifact items in maven-dependency-plugin

dwalluck: These first 8 lines or so could be deleted. It was just checking the relocation of the dependency first before checking the the same relocation was applied to the plugin/configuration sections.

PR #757, opened 2020-03-25
Ncl5283

rnc:


rnc:
@vibe13 In case you want to fix javadoc for PNC (etc) the root pom changes are I think what you need.

common/src/main/java/org/commonjava/maven/ext/common/model/Project.java
```diff
@@ -29,6 +29,7 @@
 import org.commonjava.maven.atlas.ident.ref.SimpleArtifactRef;
 import org.commonjava.maven.atlas.ident.ref.SimpleProjectVersionRef;
 import org.commonjava.maven.atlas.ident.util.VersionUtils;
+import org.commonjava.maven.ext.annotation.JavadocExclude;
```
dwalluck: Unused?

core/pom.xml
```diff
@@ -247,7 +247,7 @@
             <groupId>org.commonjava.maven.ext</groupId>
             <artifactId>pom-manipulation-annotation</artifactId>
             <classifier>processor</classifier>
-            <version>3.9-SNAPSHOT</version>
+            <version>${project.version}</version>
```
dwalluck: Weren't you saying recently not to use `${project.version}` and that maven-release-plugin would replace the version automatically anyway?

core/src/main/java/org/commonjava/maven/ext/core/groovy/ASTTransformer.java
```diff
@@ -59,17 +59,18 @@
 import static org.apache.groovy.ast.tools.ClassNodeUtils.addGeneratedMethod;
 
 /**
- * Ensures that Groovy scripts annotated with {@link } are transformed into a class that
- * extends {@link }.
+ * Ensures that Groovy scripts annotated with {@link GMEBaseScript} and {@link PMEBaseScript} are transformed into a class that
+ * extends groovy.transform.BaseScript.
```
dwalluck: Did we want to keep the `@link` for this one?

pom.xml
```diff
@@ -425,10 +428,15 @@
         <version>1.12.0</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.github.olivergondza</groupId>
+        <artifactId>maven-jdk-tools-wrapper</artifactId>
+        <version>0.1</version>
```
dwalluck: I'm confused as to why certain versions are properties and others are not.

pom.xml
```diff
@@ -585,7 +619,36 @@
           <execution>
             <id>attach-javadocs</id>
             <configuration>
+              <sourcepath>${project.build.directory}/javadoc-generated-sources</sourcepath>
+              <doclet>org.commonjava.maven.ext.annotation.JavadocFilter</doclet>
+              <docletArtifact>
+                <groupId>org.commonjava.maven.ext</groupId>
+                <artifactId>pom-manipulation-annotation</artifactId>
+                <version>${project.version}</version>
```
dwalluck: I see, because it's a dependency inside a plugin (execution/)configuration?

core/src/main/java/org/commonjava/maven/ext/core/groovy/ASTTransformer.java
rnc: Yeah, good spot.

common/src/main/java/org/commonjava/maven/ext/common/model/Project.java
rnc: Yeah! Good spot

pom.xml
rnc: That particular section is in dependencyManagement?

pom.xml
rnc: `docletArtifact` needs a full GAV

pom.xml
dwalluck: I am just curious why certain versions in `dependencyManagement` get a property defined (e.g., `slf4jVersion`), but most dependencies don't have one.

pom.xml
dwalluck: I mean is it the reason it uses `${project.version}` vs. `3.9-SNAPSHOT`.

core/pom.xml
rnc: Yes you're right, I shouldn't have changed that. Doh!

pom.xml
rnc: And for that matter, it should use a real version like 3.9-SNAPSHOT

rnc: On second thoughts after testing it - it doesn't subsitute the docletArtifact version so I won't change that.

pom.xml
rnc: If the version needs to be reused in multiple places - if you look at slf4jVersion its also used in the javadoc configuration

.gitignore
```diff
@@ -1,7 +1,4 @@
 target/
-.project
-.settings
-.classpath
 .idea
```
dwalluck: These are useful for people using Eclipse IDE.

rnc: Ah, ok.

PR #764, opened 2020-04-18
NCL-5707 Handle scope within dependencies

rnc:


rnc:
@michalszynkiewicz You might also want to take a look at this.

core/src/main/java/org/commonjava/maven/ext/core/impl/RESTCollector.java
```diff
@@ -342,9 +343,9 @@ private static void recordPlugins( ManipulationSession session, Set<ArtifactRef>
 
         for ( ProjectVersionRef pvr : plugins.keySet() )
         {
-            deps.add( new SimpleArtifactRef(
+            deps.add( new SimpleScopedArtifactRef(
                             new SimpleProjectVersionRef( pvr.asProjectRef(), handlePotentialSnapshotVersion( vs, pvr.getVersionString() ) ),
-                            new SimpleTypeAndClassifier( "maven-plugin", null ) ) );
+                            new SimpleTypeAndClassifier( "maven-plugin", null ), ArtifactScopeEnum.compile.name() ) );
```
rnc, line 348: I'm actually thinking this might be better left as null as it fits in with the defaults (empty == compile) and comparisons better.

michalszynkiewicz:
works for me :+1:

integration-test/src/it/depmgmt-strict-mode-exact-classifier/pom.xml
```diff
@@ -81,6 +81,14 @@
         <classifier>tck</classifier>
         <version>3.4.2</version>
       </dependency>
+      <dependency>
```
dwalluck, line 84: This is added twice, once with default scope (compile) and once with scope test, but scope compile implies scope test.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -209,7 +209,7 @@ private boolean updateDependencies( WildcardMap<ProjectVersionRef> relocations,
                     // Unfortunately because we iterate using the resolved project keys if the relocation updates those
                     // keys multiple iterations will not work. Therefore we need to remove the original key:dependency
                     // to map to the relocated form.
-                    postFixUp.put( SimpleArtifactRef.parse( dependency.getManagementKey() ), dependency );
+                    postFixUp.put( new SimpleScopedArtifactRef( dependency ), dependency );
```
dwalluck, line 212: Is there some reason for changing this or are the two statements equivalent? I see the parse version used later in another file.

rnc: `getManagementKey` returns something that would not be parsable - its missing a version. By using the original object that avoids potential parse issues.

integration-test/src/it/depmgmt-strict-mode-exact-classifier/pom.xml
rnc: Correct - using it to test scope and ensure that it manages to track both otherwise identical dependencies.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -369,7 +372,7 @@ private void apply( final Project project, final Model model, final Map<Artifact
                 d.setGroupId( project.getModelParent().getGroupId() );
                 d.setArtifactId( project.getModelParent().getArtifactId() );
                 d.setVersion( project.getModelParent().getVersion() );
-                pDepMap.put( SimpleArtifactRef.parse( d.getManagementKey() ), d );
+                pDepMap.put( SimpleScopedArtifactRef.parse( d.getManagementKey() ), d );
```
dwalluck, line 375: This is the reference to `SimpleScopedArtifactRef.parse( d.getManagementKey() )`.

rnc: Great spot! Fixing.

dwalluck: Looking now, I think this line is also what caused the Appveyor build failure.

rnc: That would be me making a silly mistake - fixed now

PR #765, opened 2020-04-21
NCL-5708 Expose restHeader parsing for GME. Minor code cleanup.

rnc:


rnc:
@dwalluck While verifying GME and https://github.com/project-ncl/manipulator-groovy-examples for the next release found some minor issues

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -133,4 +119,18 @@ public boolean isRestSuffixAlign()
     {
         return restSuffixAlign;
     }
+
+    // Public API : Used by GME to convert as well.
+    public static Map<String,String> restHeaderParser(String value)
```
rnc, line 124: @dwalluck Moving this to a function means we can reuse it easily in GME for its configuration as well.

.appveyor.yml
```diff
@@ -34,7 +34,7 @@ cache:
   - C:\Users\appveyor\.m2 -> pom.xml
 
 build_script:
-  - ps: mvn -B -Prun-its clean install
```
rnc, line 37: We kept hitting the limit for appveyor and the build was timing out - by removing some checks/generation hopefully it will now pass.

PR #773, opened 2020-05-01
NCL5723 - Preserve formatting changes when writing POM files. 

### LINKED ISSUE #637: PME should minimize formatting changes to the files
When PME changes the files it also changes the formatting elements for some items. This makes reviewing PME changes harder, because it tends to change items that are unrelated to the actual changes in place. Here's a couple of examples taken from a diff of ours:

Before PME:
```
<?xml version="1.0"?>
```

After PME:
```
<?xml version="1.0" encoding="UTF-8"?>
```

Before PME:
```
<Import-Package>
  javax.ws.rs.*;version="[1.1,3)",
  org.apache.cxf.*;version="[2,4)",
</Import-Package>
```

After PME:
```
<Import-Package>javax.ws.rs.*;version="[1.1,3)",
    org.apache.cxf.*;version="[2,4)",</Import-Package>
<Export-Package>org.apache.cxf.samples.failover.jaxws</Export-Package>
```

Before PME: 
```<ignore />```

After PME:
```<ignore></ignore>```

We would like to request that PME refrain from doing these type of changes or, at least, provide a way that users can control how it will handle formatting of the elements it touches.

### PR DESCRIPTION
**Pull request by rnc**
@jdcasey For reference this removes the PME dependency on https://github.com/Commonjava/maven-model-jdom-support/ 

Fixes #637

### CODE REVIEW

**Comment by rnc** [comment]
@dwalluck Any idea why the coverage results haven't appeared?

**Comment by dwalluck** [comment]
> @dwalluck Any idea why the coverage results haven't appeared?

If I login to codecov.io, I think I see the results. You could try to re-request access to this repo (I think only an owner can do it).

**Comment by rnc** [comment]
@dwalluck Can you review this please

File: core/src/test/java/org/commonjava/maven/ext/core/impl/InitialGroovyManipulatorTest.java [hunk 1]
```diff
@@ -84,6 +85,7 @@ public void shouldRemoveProjectInGroovyScript() throws Exception
         projects = pomIO.parseProject( projectroot );
         assertThat( projects.size(), equalTo( 3 ) );
 
+        assertEquals( projectForArtifactId( projects, "groovy-project-removal").getModel().getProfiles().size(), 1 );
```
dwalluck, line 88: Arguments here are reversed: should be expected, actual

io/src/test/java/org/commonjava/maven/ext/io/XMLIOTest.java
```diff
@@ -212,6 +212,15 @@ public void modifyNotFoundFile ()
         assertEquals( 0, nodeList.getLength() );
     }
 
+    @Test
+    public void findArtemis() throws ManipulationException, XPathExpressionException
+    {
+        Document doc = xmlIO.parseXML( xmlFile );
+        XPath xPath = XPathFactory.newInstance().newXPath();
+        NodeList nodeList = (NodeList) xPath.evaluate( "//include[starts-with(.,\"org.apache.tomcat\")]", doc, XPathConstants.NODESET );
```
dwalluck, line 220: I am not really sure I understand why the method is called `findArtemis` but the tests looks for `org.apache.tomcat`.

common/src/main/java/org/commonjava/maven/ext/common/util/LineSeparator.java
```diff
@@ -0,0 +1,60 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.commonjava.maven.ext.common.util;
+
+/**
+ * Partially taken from org.jdom2.output.LineSeparator as the Maven Release Plugin uses JDOM1 not 2.
+ */
```
dwalluck, line 21: Should the file have the original copyright/license header then?

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
```diff
@@ -140,4 +201,31 @@ public void testWriteModel()
...
+        FileUtils.copyFile( pom, targetFile );
+
+        List<Project> projects = pomIO.parseProject( targetFile );
+        projects.get( 0 ).setExecutionRoot();
+
+        HashSet<Project> changed = new HashSet<>();
+        changed.add( projects.get( 0 ) );
+        pomIO.rewritePOMs( changed );
+
+        String s = FileUtils.readFileToString( targetFile, Charset.defaultCharset() );
```
dwalluck, line 224: Would be more accurate if you could get the charset from the pom: `Charset.forName( projects.get( 0 ).getModel().getModelEncoding() )`.

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
```diff
@@ -140,4 +201,31 @@ public void testWriteModel()
...
+        assertEquals( StringUtils.countMatches(s, "Modified by POM Manipulation Extension" ), 1);
+
+        pomIO.rewritePOMs( changed );
+        s = FileUtils.readFileToString( targetFile, Charset.defaultCharset() );
```
dwalluck, line 228: Same comment as above.

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
```diff
@@ -140,4 +201,31 @@ public void testWriteModel()
...
+        assertEquals( StringUtils.countMatches(s, "Modified by POM Manipulation Extension" ), 1);
```
dwalluck, line 229: Arguments here are reversed. Should be: expected, actual.

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
```diff
@@ -70,7 +105,6 @@ public void testRewritePOMs()
         Project p = new Project( targetFile, model );
         HashSet<Project> changed = new HashSet<>();
         changed.add( p );
-
         pomIO.rewritePOMs( changed );
 
         assertTrue( FileUtils.contentEqualsIgnoreEOL( pom, targetFile, StandardCharsets.UTF_8.toString() ) );
```
dwalluck, line 110: Charset could be `Charset.forName( model.getModelEncoding() ) `;

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
```diff
@@ -48,6 +56,29 @@ public void setup()
...
+        File targetFile = folder.newFile( "target.xml" );
+        FileUtils.copyFile( pom, targetFile );
+
+        List<Project> projects = pomIO.parseProject( targetFile );
+        // We don't want this to be the execution root so that it doesn't add "Modified by" which breaks the comparison
+        FieldUtils.writeDeclaredField( projects.get( 0 ), "executionRoot", false, true);
+        HashSet<Project> changed = new HashSet<>(projects);
+        pomIO.rewritePOMs( changed );
+
+        assertTrue( FileUtils.contentEqualsIgnoreEOL( pom, targetFile, StandardCharsets.UTF_8.toString() ) );
```
dwalluck, line 78: Charset could be `Charset.forName(  projects.get( 0 ).getModel().getModelEncoding() )`.

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: Ok

io/src/test/java/org/commonjava/maven/ext/io/XMLIOTest.java
rnc: Probably because its looking at an xml file called `activemq-artemis`. :-) I'll rename it

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: Ok

common/src/main/java/org/commonjava/maven/ext/common/util/LineSeparator.java
rnc: Ok

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: I've moved everything to using StandardCharsets.UTF_8 as the model either has that or null (which breaks the tests)

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: I've moved everything to using StandardCharsets.UTF_8 as the model either has that or null (which breaks the tests)

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: I've moved everything to using StandardCharsets.UTF_8 as the model either has that or null (which breaks the tests)

io/src/test/java/org/commonjava/maven/ext/io/PomIOTest.java
rnc: I've moved everything to using StandardCharsets.UTF_8 as the model either has that or null (which breaks the tests)

io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java
```diff
@@ -57,7 +58,7 @@ public void write( Settings settings, File settingsFile )
     {
         try
         {
-            PrintWriter printWriter = new PrintWriter( settingsFile, settings.getModelEncoding() );
+            PrintWriter printWriter = new PrintWriter( settingsFile, StandardCharsets.UTF_8.toString() );
```
dwalluck, line 61: Why change here? You mentioned partial models in test as a reason, but this is not a test case. This file could have a different encoding set by the user. Recommend:

```
settings.getModelEncoding() != null ? settings.getModelEncoding() : StandardCarsets_UTF_8.name();
```

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
```diff
@@ -248,49 +257,63 @@ private void write( final Project project, final File pom, final Model model )
...
+                    outtro = commentStart + manifestComment + commentEnd;
                 }
-            });
-        }
-        catch ( final IOException e )
-        {
-            throw new ManipulationException( "Failed to read POM for rewrite: {}. Reason: ", pom, e.getMessage(), e );
+                else
+                {
+                    outtro = outtro.replaceAll( "Modified by.*", manifestComment );
```
dwalluck, line 306: Probably this could use the `MODIFIED_BY + .*` constant that you already have defined.

rnc: I don't think so - the modified by includes a version and we want to ensure we replace it by the correct version. Its feasible it could have an older version.

io/src/main/java/org/commonjava/maven/ext/io/SettingsIO.java
rnc: Agreed. ( looked at 4.3.3 in https://www.w3.org/TR/xml/ )

PR #777, opened 2020-06-25
NCL-5745: Fix docIndex for overrideTransitive config value

dwalluck:


rnc:
@dwalluck Is there a way we can make those coverage complaints less picky?

dwalluck:
> @dwalluck Is there a way we can make those coverage complaints less picky?

I have tried to set it to 1% with #779, but I guess it has to be committed first to test it.

PR #778, opened 2020-06-25
NCL-5745: Use same line separator and charset across all platforms

dwalluck:


dwalluck:
@rnc Codecov passes now

rnc:
Nice!

annotation/src/main/java/org/commonjava/maven/ext/annotation/ConfigValueProcessor.java
```diff
@@ -141,12 +141,12 @@ void generateCode() throws IOException
                 .append( value )
```
rnc, line 141: Did you want to deal with the \n elsewhere?

dwalluck: I am not sure what you mean. This is supposed to make the generated `target/property-index.md` file always have UNIX newlines, so that no matter what OS you run the build on you get the same exact file.

dwalluck: Not to mention that there is already a line with hard-coded `\n`:

```
propertyIndex.append( "---\ntitle: \"Index of Properties\"\n---\n\n" );
````

which was creating lines with both `\n` and `\r\n` on Windows. Now, there is just `\n` in the file.

rnc: OK cool

PR #779, opened 2020-06-26
Set codecov project and patch threshold

dwalluck:


rnc:
@dwalluck  This seems to be going very slow and not reporting back (travis has succeeded at time of writing) - shall we just merge this?

PR #782, opened 2020-07-12
NCL-5715 Attempt to run some tests in parallel. Use slf4j-test to avo…

rnc:
…id problems with parallel tests. Refactor test code. Run invoker tests in parallel. Fix javadoc issues.


integration-test/src/test/java/org/commonjava/maven/ext/integrationtest/DefaultCliIntegrationTest.java
```diff
@@ -33,11 +33,11 @@
 import static org.commonjava.maven.ext.integrationtest.ITestUtils.runLikeInvoker;
 import static org.commonjava.maven.ext.integrationtest.ITestUtils.runMaven;
 
-@SuppressWarnings( "ConstantConditions" )
 @RunWith( Parameterized.class )
 public class DefaultCliIntegrationTest
 {
     private static final Logger LOGGER = LoggerFactory.getLogger( DefaultCliIntegrationTest.class );
+    private static final Logger logger = LoggerFactory.getLogger( DefaultCliIntegrationTest.class );
```
dwalluck, line 40: Looks like we've got two loggers here.

io/src/test/java/org/commonjava/maven/ext/io/rest/VersionTranslatorTest.java
```diff
@@ -176,19 +166,8 @@ public void testTranslateVersionsFailNoResponse()
...
-        {
-            // Disable logging for this test as impacts timing.
-            logbackLogger.setLevel( Level.OFF );
-            versionTranslator.translateVersions( aLotOfGavs );
-        }
-        finally
-        {
-            logbackLogger.setLevel( originalLevel );
-        }
+        // Disable logging for this test as impacts timing.
```
dwalluck, line 169: Comment is still here,  but it looks like the logger code was removed.

cli/pom.xml
```diff
@@ -182,6 +182,13 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration combine.self="override">
```
dwalluck, line 188: What issue does this `combine.self` solve?

integration-test/src/test/java/org/commonjava/maven/ext/integrationtest/DefaultCliIntegrationTest.java
rnc: Great spot!

io/src/test/java/org/commonjava/maven/ext/io/rest/VersionTranslatorTest.java
rnc: Great spot!

cli/pom.xml
rnc: Preventing overriding the parent configuration (See https://maven.apache.org/pom.html ) - I think, if I recall correctly, the CLITest was not suitable for forking etc

dwalluck: Do you mean the parent from overriding the child configuration? I think there is `<inherited>false</inherited>` to prevent merging the configuration with the child. I see `inherited` used in a lot of the examples in that URL, but nowhere in the text do they discuss that option.

I wasn't familiar with this option before, so a comment would have been helpful to me.

rnc: Ah ok. Its the inverse of `<inherited>false` as we normally want the inheritance except for this module. Can go over it offline if you want.

rnc:
Closing this, as with the migration to GitHub CI its not a priority and did not make such a significant time saving on CI runs.

PR #788, opened 2020-09-22
NCLSUP64 Modify RelocationManipulator so it runs before REST phase so…

rnc:
… its subject to alignment


rnc:
@dwalluck Feedback welcome :-)

PR #792, opened 2020-10-06
NCL-6051 Port to picocli. Add option to output manipulator order

rnc:


rnc:
Will need to be updated and rebased once #783 is merged.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -125,99 +148,23 @@ public static void main( String[] args )
...
-                                 .hasArgs()
-                                 .numberOfArgs( 1 )
-                                 .desc( "Optional settings.xml file" )
-                                 .build() );
-        options.addOption( Option.builder( "P" )
-                                 .longOpt( "activeProfiles" )
-                                 .desc( "Comma separated list of active profiles." )
-                                 .numberOfArgs( 1 )
-                                 .build() );
-        options.addOption( Option.builder( "o" )
```
dwalluck, line 157: Seems we lost this option?

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -125,99 +148,23 @@ public static void main( String[] args )
...
-                                 .longOpt( "outputFile" )
-                                 .desc( "outputFile to output dependencies to. Only used with '-p' (Print all project dependencies)" )
-                                 .numberOfArgs( 1 )
-                                 .build() );
-        options.addOption( Option.builder( "p" ).longOpt( "printDeps" ).desc( "Print all project dependencies" ).build() );
```
dwalluck, line 162: Seems you changed this option name, but you also missed adding short option `-p`.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -125,99 +148,23 @@ public static void main( String[] args )
...
-        options.addOption( Option.builder()
```
dwalluck, line 163: This is missing.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -125,99 +148,23 @@ public static void main( String[] args )
...
-                                 .longOpt( "printGAVTC" )
-                                 .desc( "Print all project dependencies as group:artifact:version:type:classifier" )
-                                 .build() );
-        options.addOption( Option.builder( "D" )
-                                 .hasArgs()
-                                 .numberOfArgs( 2 )
-                                 .valueSeparator( '=' )
-                                 .desc( "Java Properties" )
-                                 .build() );
-        options.addOption( Option.builder( "x" )
```
dwalluck, line 173: This is missing.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -260,7 +207,7 @@ public int run( String[] args )
             return 10;
         }
         // Don't bother skipping if we're just trying to analyse deps.
-        else if ( new File( target.getParentFile(), ManipulationManager.MARKER_FILE ).exists() && !cmd.hasOption( 'p' ) )
+        else if ( new File( target.getParentFile(), ManipulationManager.MARKER_FILE ).exists() )
```
dwalluck, line 210: The previous code also checked `!printDeps` which I guess is `!printProjectDeps` now? I assume we want to add an alias for the old option names if we change them here?

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
```diff
@@ -185,83 +187,210 @@ public void checkLocalRepositoryWithDefaultsAndModifiedUserSettings() throws Exc
...
+
+    @Test
+    public void checkDependencies()
+    {
+        final File root = Paths.get( ROOT_DIRECTORY.toString(), "pom.xml" ).toFile();
+
+        expectedSystemExit.expectSystemExitWithStatus( 0 );
+        Cli.main( new String[] { "--printProjectDeps", "--file", root.getAbsolutePath() } );
+
+        assertThat( systemOutRule.getLog(), equalTo( "INFO org.commonjava.maven.ext.cli.Cli - Found 82 dependencies\n" + "ch.qos.logback:logback-classic:1.2.3                                            jar                 compile             \n"
```
dwalluck, line 279: There is a newline issue here since `System.out.println()` line separator varies.

You could just use `getLogWithNormalizedLineSeparator()` instead of `getLog()` here.

It would be more accurate to change `\n` to `String.format()` with `%n` because that would verify the "correct" line separator vs. ignoring it.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -111,6 +113,7 @@ public ManipulationManager( Map<String, Manipulator> manipulators,
     /**
      * Determined from {@link Manipulator#getExecutionIndex()} comparisons during {@link #init(ManipulationSession)}.
      */
+    @Getter
```
dwalluck, line 116: Sometimes it's better/safer to return `Collections.unmodifibleList()` if it's intended to be read-only rather than a direct reference and lombok could really make things easier if it supported such an option, but it doesn't, see <https://github.com/rzwitserloot/lombok/issues/1504>.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/ITestUtils.java
```diff
@@ -196,7 +198,10 @@ static void runLikeInvoker( String workingDir, String url )
     static Integer runCli( List<String> args, Map<String, String> params, String workingDir )
     {
         ArrayList<String> arguments = new ArrayList<>( args );
-        Collections.addAll( arguments, toJavaParams( params ).split( "\\s+" ) );
+        if (params.size() > 0 )
```
dwalluck, line 201: Very very minor, but I prefer `!isEmpty()` to `size() > 0`. I checked `HashMap` and it literally checks `size == 0`, so they are the essentially the same, but for some things, you could imagine they have different performance. Also, logically I think it makes more sense as to what is actually being tested.

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
```diff
@@ -121,18 +124,16 @@ public void checkTargetDefaultMatches() throws Exception
...
-        MavenSession ms = (MavenSession)FieldUtils.readField( session, "mavenSession", true );
+        MavenSession ms = (MavenSession) FieldUtils.readField( session, "mavenSession", true );
 
-        assertEquals( ms.getRequest().getLocalRepository().getBasedir(),
-                      ms.getRequest().getLocalRepositoryPath().toString() );
+        assertEquals( ms.getRequest().getLocalRepository().getBasedir(), ms.getRequest().getLocalRepositoryPath().toString() );
         assertEquals( "File " + new File( ms.getRequest().getLocalRepository().getBasedir() ).getParentFile().toString()
-                                      + " was not equal to " + System.getProperty( "user.home" ) + File.separatorChar
-                                      + ".m2",
+                                      + " was not equal to " + System.getProperty( "user.home" ) + File.separatorChar + ".m2",
```
dwalluck, line 136: Minor, but why does the code use `separatorChar` instead of `separator` if it's concatenating strings?

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
```diff
@@ -142,17 +143,18 @@ public void checkLocalRepositoryWithDefaults() throws Exception
...
         Path backup = Paths.get( source.toString() + '.' + UUID.randomUUID().toString() );
-        Path tmpSettings = Paths.get ( getClass().getResource("/settings-test.xml").toURI() );
+        Path tmpSettings = Paths.get( getClass().getResource( "/settings-test.xml" ).toURI() );
 
         try
         {
             if ( source.toFile().exists() )
             {
-                System.out.println ("Backing up settings.xml to " + backup);
+                System.out.println( "Backing up settings.xml to " + backup );
```
dwalluck, line 155: Logger should be used.

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
```diff
@@ -185,83 +187,210 @@ public void checkLocalRepositoryWithDefaultsAndModifiedUserSettings() throws Exc
...
+                                                                     + "org.projectlombok:lombok-maven-plugin:1.18.12.0                                 maven-plugin                            \n"
+                                                                     + "org.slf4j:slf4j-api:1.7.30                                                      jar                 compile             \n"
+                                                                     + "org.xmlunit:xmlunit-core:2.1.1                                                  jar                 test                \n"
+                                                                     + "org.xmlunit:xmlunit-matchers:2.1.1                                              jar                 test                \n"
+                                                                     + "org.yaml:snakeyaml:1.17                                                         jar                 compile             \n" ) );
+
+    }
+
+    @Test
+    public void checkManipulatorOrder()
```
dwalluck, line 365: It makes this test a little hard to update when adding a new manipulator.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Yes :-)

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Yes :-)

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: On purpose

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: On purpose

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
rnc: Well, its kind of on purpose :-)

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
rnc: Its partly because I copy and paste the text in intellij and it adds the separators. I can use `getLogWithNormalizedLineSeparator`.

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
dwalluck: I realize it has to be hardcoded, but it depends if this test is really is meant to test the particular order or just that the output works.

If you wanted to verify the relative ordering of the manipulators, I assume that is another test, and if you just want to verify that output works, it could (potentially) be made less strict. But, it's not a big deal.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
dwalluck: I guess this obviously needs accompanying documentation listed the removed options.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
dwalluck: But do you want to add a short-option `-p` to the new replacement option as well?

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Good point.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Ok, that seems reasonable. Will change it.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Done in https://github.com/release-engineering/pom-manipulation-ext/pull/794

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
rnc: Its an interesting point, but I'm not sure its really worth it.

integration-test/src/main/java/org/commonjava/maven/ext/integrationtest/ITestUtils.java
rnc: Done

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
rnc: Doesn't really matter as its in a test and the output is muted anyway inside the systemoutrule.

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
rnc: Shrug .... yeah probably could be change but this was pre-existing code.

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
rnc: I think a bit of both - if we accidentally changed the order this would catch it, and this also tests the output.

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Done

cli/src/test/java/org/commonjava/maven/ext/cli/CliTest.java
dwalluck: Muted? So where can you see the message at all, then?

rnc: If the test fails :-)

rnc: I could just remove it, but it was preexisting code and I hadn't changed it.

dwalluck (approved):

PR #796, opened 2020-10-07
Add classifier and colour to dependency output. Make tests pme versio…

rnc:
…n independent

### DIFF
--- a/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
+++ b/cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
@@ -256,23 +256,26 @@ public class Cli implements Callable<Integer>
             {
                 List<ArtifactRef> ts = RESTCollector.establishAllDependencies( session, pomIO.parseProject( session.getPom() ),
                                                                               profiles ).stream().sorted().collect(Collectors.toList());
-                logger.info( "Found {} dependencies", ts.size());
+                System.out.format( "Found %s dependencies\n", ts.size());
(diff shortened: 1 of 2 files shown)


cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -256,23 +256,26 @@ else if ( new File( target.getParentFile(), ManipulationManager.MARKER_FILE ).ex
...
                                           a.getTypeAndClassifier().getType(),
+                                          a.getTypeAndClassifier().getClassifier() == null ? "" : a.getTypeAndClassifier().getClassifier(),
                                           isPlugin ? "" : scope);
                     }
                 }
             }
             else if (printManipulatorOrder)
             {
-                System.out.println ("Manipulator order is:");
+                System.out.println ("\u001B[32mManipulator order is:\033[0m");
```
dwalluck, line 278: I was thinking maybe we had Jansi to help with this, but it's only in test scope `org.fusesource.jansi:jansi:jar:1.13:test`.

dwalluck: I'm guessing if we wait for picocli to go int hen we could use `org.fusesource.jansi.AnsiConsole`.

rnc: It seems excessive to add jansi for two lines of code.

PR #808, opened 2020-12-01
NCLSUP-209 Detect running in container to disable log file

rnc:


vibe13 (approved):
Looks good to me Nick! Thanks for this improvement!

PR #820, opened 2021-01-10
[NCL-6166] Add build type flags for DA calls

pkocandr:
Introduce command line flag restBrewPullActive and new switch restMode. Those are used to determine if the REST service (DA) should use the Brew Pull feature or not and if it should also lookup temporary and/or managed service artifacts or not. This should replace passing of restRepositoryGroup value and the repository group to be used should be determined by DA based on the flags. All of those command line params are passed to the rest service which should ensure it's backwards compatible.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
@@ -41,6 +41,12 @@ public class RESTState implements State
     public static final String REST_REPO_GROUP = "restRepositoryGroup";
 
     @ConfigValue( docIndex = "dep-manip.html#rest-endpoint" )
+    public static final String REST_BREW_PULL_ACTIVE = "restBrewPullActive";
+
+    @ConfigValue( docIndex = "dep-manip.html#rest-endpoint" )
+    public static final String REST_MODE = "restMode";
+
+    @ConfigValue( docIndex = "dep-manip.html#rest-endpoint" )
     public static final String REST_MAX_SIZE = "restMaxSize";
 
     @ConfigValue( docIndex = "dep-manip.html#rest-endpoint" )
@@ -76,6 +82,7 @@ public class RESTState implements State
         initialise( session.getUserProperties() );
     }
 
+    @Override
     public void initialise( Properties userProps )
     {
         final VersioningState vState = session.getState( VersioningState.class );
@@ -84,6 +91,8 @@ public class RESTState implements State
         restSuffixAlign = Boolean.parseBoolean( userProps.getProperty( REST_SUFFIX, "true" ) );
 
         String repositoryGroup = userProps.getProperty( REST_REPO_GROUP, "" );
+        Boolean brewPullActive = Boolean.parseBoolean( userProps.getProperty( REST_BREW_PULL_ACTIVE ) );
+        String mode = userProps.getProperty( REST_MODE );
         int restMaxSize = Integer.parseInt( userProps.getProperty( REST_MAX_SIZE, "-1" ) );
         int restMinSize = Integer.parseInt( userProps.getProperty( REST_MIN_SIZE,
                                                                    String.valueOf( DefaultTranslator.CHUNK_SPLIT_COUNT ) ) );
@@ -95,8 +104,9 @@ public class RESTState implements State
         int restRetryDuration = Integer.parseInt( userProps.getProperty( REST_RETRY_DURATION_SEC,
                                                                          String.valueOf( DefaultTranslator.RETRY_DURATION_SEC ) ) );
 
-        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, vState.getIncrementalSerialSuffix(),
-                                              restHeaders, restConnectionTimeout, restSocketTimeout, restRetryDuration );
+        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, brewPullActive, mode,
+                                              vState.getIncrementalSerialSuffix(), restHeaders, restConnectionTimeout,
+                                              restSocketTimeout, restRetryDuration );
     }
 
     /**
--- a/io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
+++ b/io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
@@ -76,6 +76,10 @@ public class DefaultTranslator
 
     private final String repositoryGroup;
 
+    private final Boolean brewPullActive;
+
+    private final String mode;
+
     private final String incrementalSerialSuffix;
 
     private final Map<String, String> restHeaders;
@@ -101,15 +105,21 @@ public class DefaultTranslator
      * @param endpointUrl is the URL to talk to.
      * @param restMaxSize initial (maximum) size of the rest call; if zero will send everything.
      * @param restMinSize minimum size for the call
-     * @param repositoryGroup the group to pass to the endpoint.
+     * @param repositoryGroup the group to pass to the endpoint, still can be used, but is replaced by
+     *        brewPullActive flag and mode when used in combination with DA 2.1 and above
+     * @param brewPullActive flag saying if brew pull should be used for version retrieval
+     * @param mode lookup mode, either STANDARD (default if empty) or SERVICE
      * @param incrementalSerialSuffix the suffix to pass to the endpoint.
      * @param restHeaders the headers to pass to the endpoint
      */
     public DefaultTranslator( String endpointUrl, int restMaxSize, int restMinSize, String repositoryGroup,
-                              String incrementalSerialSuffix, Map<String, String> restHeaders,
-                              int restConnectionTimeout, int restSocketTimeout, int restRetryDuration )
+                              Boolean brewPullActive, String mode, String incrementalSerialSuffix,
+                              Map<String, String> restHeaders, int restConnectionTimeout, int restSocketTimeout,
+                              int restRetryDuration )
     {
         this.repositoryGroup = repositoryGroup;
+        this.brewPullActive = brewPullActive;
+        this.mode = mode;
         this.incrementalSerialSuffix = incrementalSerialSuffix;
         this.endpointUrl = endpointUrl + ( isNotBlank( endpointUrl ) ? endpointUrl.endsWith( "/" ) ? "" : "/" : "");
         this.initialRestMaxSize = restMaxSize;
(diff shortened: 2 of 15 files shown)


pkocandr:
One known downside of this is that also the built projects modules' version is determined using these flags. In context of PNC the brewPullActive should always be true for them to ensure it doesn't rebuild an already existing version built in Brew. It would fail later in promotion validation, but it is too late.
To fix this in case restBrewPullActive=false is passed we should split the request to DA to check the latest versions of built modules with brewPullActive=true and versions of dependencies should used the passed value (brewPullActive=false).

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -95,8 +105,9 @@ public void initialise( Properties userProps )
         int restRetryDuration = Integer.parseInt( userProps.getProperty( REST_RETRY_DURATION_SEC,
                                                                          String.valueOf( DefaultTranslator.RETRY_DURATION_SEC ) ) );
 
-        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, vState.getIncrementalSerialSuffix(),
-                                              restHeaders, restConnectionTimeout, restSocketTimeout, restRetryDuration );
+        restEndpoint = new DefaultTranslator( restURL, restMaxSize, restMinSize, repositoryGroup, brewPullActive, temporaryBuild,
```
rnc, line 108: Question - if `brewPullAction` and `temporaryBuild` are replacing `repositoryGroup` why do we still have `repositoryGroup` here?

rnc:
Aside - I presume you would also be doing a matched PR for the documentation?

pom.xml
```diff
@@ -78,6 +78,7 @@
 
     <mavenVersion>3.5.0</mavenVersion>
     <atlasVersion>0.17.1</atlasVersion>
+    <daVersion>2.1.0-SNAPSHOT</daVersion>
```
rnc, line 81: Is there not a non-snapshot version to use?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
pkocandr: For backwards compatibility so we can deploy new PME independently on DA/Repour. We can either use the new PME version earlier than them or later.

pom.xml
pkocandr: Not yet. I was thinking about separating the API from DA and release it once finalized for PNC 2.1 even though we will be still in development phase. Have to discuss that with the team.

pkocandr:
> Aside - I presume you would also be doing a matched PR for the documentation?

Right. I need to take a look into it too. Thanks for reminding me.

pom.xml
rnc: OK, we will need to discuss this because right now it didn't even get to the tests as it isn't picking up the snapshot.

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -84,6 +92,8 @@ public void initialise( Properties userProps )
         restSuffixAlign = Boolean.parseBoolean( userProps.getProperty( REST_SUFFIX, "true" ) );
 
         String repositoryGroup = userProps.getProperty( REST_REPO_GROUP, "" );
+        Boolean brewPullActive = BooleanUtils.toBooleanObject( userProps.getProperty( REST_BREW_PULL_ACTIVE ) );
+        Boolean temporaryBuild = BooleanUtils.toBooleanObject( userProps.getProperty( REST_TEMPORARY_BUILD ) );
```
rnc, line 96: Why not use `Boolean.parseBoolean` rather than an external library?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
rnc: Ok makes sense. Couple of options I guess - I release 4.3 now, and then this becomes 4.4 with `repositoryGroup` removed or we keep it and release in with 4.3 but mark it as deprecated via the `ConfigValue` annotation )with appropriate updates in the documentation) and then remove it in future.

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
pkocandr: Because the logic using the flags also need to know if the value was passed in or not (using BooleanUtils.toBooleanObject(null) results in null). There is no such method in Boolean class AFAIK.

rnc: Huh, is the boolean being used as a tri-state value ?

pom.xml
```diff
@@ -78,6 +78,7 @@
 
     <mavenVersion>3.5.0</mavenVersion>
     <atlasVersion>0.17.1</atlasVersion>
+    <daVersion>2.1.0.Api-1</daVersion>
```
rnc, line 81: Wow, what a GAV!! :-)

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -84,6 +92,8 @@ public void initialise( Properties userProps )
         restSuffixAlign = Boolean.parseBoolean( userProps.getProperty( REST_SUFFIX, "true" ) );
 
         String repositoryGroup = userProps.getProperty( REST_REPO_GROUP, "" );
+        Boolean brewPullActive = BooleanUtils.toBooleanObject( userProps.getProperty( REST_BREW_PULL_ACTIVE ) );
```
rnc, line 95: Why can't we use `Boolean.parseBoolean` here?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -101,15 +105,21 @@
      * @param endpointUrl is the URL to talk to.
      * @param restMaxSize initial (maximum) size of the rest call; if zero will send everything.
      * @param restMinSize minimum size for the call
-     * @param repositoryGroup the group to pass to the endpoint.
+     * @param repositoryGroup the group to pass to the endpoint, still can be used, but is replaced by
+     *        brewPullActive flag and mode when used in combination with DA 2.1 and above
+     * @param brewPullActive flag saying if brew pull should be used for version retrieval
```
rnc, line 110: So just to clarify, `repositoryGroup` isn't deprecated as such - as its still in use, until the next PNC release when it can be removed? (i.e. we'll need a separate ticket to remind us to remove the flag for that?)

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -101,15 +105,21 @@
...
+     * @param mode lookup mode, either STANDARD (default if empty) or SERVICE
```
rnc, line 111: What happens if someone passes something else (e.g. `foobar`) to the flag? Should it be an enum / or have some other checks? Or is there something in DA we can call on to validate it?

rnc: Also what does it do? Should we have test to validate it?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
rnc: Marking as resolved - seems sensible to have a release with backwards compatibility

pom.xml
pkocandr: Yeah, we needed to make a partial release because of a snapshot transitive dependency we cannot release ATM, so to emphisize that we chose this version suffix.

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
pkocandr: Boolean.parseBoolean produces boolean, not Boolean, i.e. the null value (when not provided on command line) gets lost in favor of a default one. But the default should be decided on DA side, not in PME.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
pkocandr: Yeah, we introduced `brewPullActive` and `mode` which tells DA which repository group to use (if any). The `repositoryGroup` parameter is meant to be used in combination with DA 2.0 and with DA 2.1 we will switch to use the other 2.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
pkocandr: The modes are dynamically configured in DA as needed.  Currently we plan to use 4 - PERSISTENT, TEMPORARY, SERVICE and SERVICE-TEMPORARY. It specifies allowed version suffixes, build categories and artifact qualities when looking up existing versions. If somebody passes a non-existing mode, it will fail in DA.

pom.xml
rnc: Ok!

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Ok thanks! Can another ticket be spawned to note removal of `repositoryGroup` please?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
rnc: Won't it get serialized via jackson when calling DA? So it would end up as false/true ?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: So, inside PNC would repour pass the correct mode to pme, which passes it through to DA?

And the mode isn't hardcoded in DA, but configured on service start? Should pme handle non-existing mode specially ?

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
pkocandr: It depends on the serialization implementation, but I expected that null value will be skipped from putting into the resulting json.
But anyway, the exptra fields are ignored and in DA 2.1 these become mandatory, so sending in a value is probably ok. Will update it to use `Boolean.parseBoolean()`.

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
pkocandr: Created as NCL-6528. Thank you!

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
pkocandr: Yes, Repour is responsible to pass the mode based on the build type etc. and the modes are configured in both DA and Repour at service start.
If a mode doesn't exist it should be returned by DA in an error response. If used directly by users they should be able to read it from the log (I expect the exception from the response should be printed in the terminal) and in case of running in PNC the sustaining team will find it in the logs to understand the problem and fix the config.
What kind of special handling would you suggest?

rnc: None really :-) Let me rephrase, if an erroneous response is printed sufficiently clearly for people to understand, thats fine. I was just wondering if the error response should be parsed and in this case made really explicit what has gone wrong. But if you think its fine as is, thats fine.

pkocandr: Actually when I tested it now it failed with `REST communication with http://da-master-devel.psi.redhat.com/da/rest/v-1 failed. Received response status -1 with message: Problem in HTTP communication with status code 500 and message Internal Server Error`. This is definitely suboptimal, but must be handled on DA side IMO.

rnc (approved):

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Sure, we can always do another iteration I guess if PME needs further changes

pkocandr: I've added one more commit to handle this. It turned out the exception is in the response body, but PME did not print it even with debug on. Now it is printed like this:
```ERROR org.commonjava.maven.ext.cli.Cli - REST communication with http://da.example.com/da/rest/v-1 failed. Received response status 500 with message: javax.validation.ValidationException: Invalid mode name: a. Available modes are: [PERSISTENT, TEMPORARY]```

PR #824, opened 2021-04-13
[NCL-6166] Add build type flags for DA calls - documentation

pkocandr:


pkocandr:
This is docs update for #820.

guide/dep-manip.md
```diff
@@ -40,7 +40,11 @@ PME will then call the following endpoints
...
+It will initially call the `lookup/gavs` endpoint. By default PME will pass *all* the GAVs to the endpoint **automatically auto-sizing** the data sent to DA according to the project size. Note that the initial split batches can also be configured manually via `-DrestMaxSize=<...>`. If that value is set to 0, then everything is sent without any auto-sizing. If the endpoint returns a 503 or 504 timeout the batch is automatically split into smaller chunks in an attempt to reduce load on the endpoint and the request retried. It will by default chunk down to size of 4 before aborting. This can be configured with `-DrestMinSize=<...>`.
+
+An optional `restRepositoryGroup` parameter may be specified so that the endpoint can use a particular repository group. This parameter is meant to be used with DA versions up to 2.0.
```
rnc, line 45: Nit : acronym `DA` is being used but it hasn't been explained - I suggest modifying line 34 to add the acronym there.

guide/dep-manip.md
```diff
@@ -40,7 +40,11 @@ PME will then call the following endpoints
...
+
+With DA 2.1 it is being replaced with a boolean flag `restBrewPullActive` and a string identifier `restMode`. The `restBrewPullActive` flag switches on and off the version lookup in Brew. Switching it off might have positive effect on performance. The `restMode` indicates type of versions to lookup. Modes are configurable in DA, so it is needed to check the DA config/consult with DA maintainers for the list of available modes.
```
rnc, line 48: Will there be links in DA documentation to the modes? I think having (example?) of https://github.com/release-engineering/pom-manipulation-ext/pull/820#discussion_r612347816 would be useful

guide/dep-manip.md
pkocandr: I was thinking about the same, but since it was already used before my changes ("the data sent to DA according to the project size") I expected it is ok.

guide/dep-manip.md
pkocandr: I would expect the better place for that to be somewhere in PNC User's Guide, which covers the deployment specific confgiuration. Putting it here we would be forced to maintain it with future chenges of the configuration.

rnc: The user's guide is private so its not suitable.

guide/dep-manip.md
rnc: True, lets use this opportunity to fix it :-)

pkocandr: OK, added an explanation in there.

guide/dep-manip.md
pkocandr: Yeah, but the configuration is also private. That's why it makes sense to me.

pkocandr: Anyway, I added the basic examples there.

rnc (approved):

PR #825, opened 2021-04-15
NCL-6081: Change alignment report filename and make output more consi…

dwalluck:
…stent

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
@@ -80,7 +80,7 @@ public class ManipulationManager
 
     public static final String MARKER_FILE = MARKER_PATH + File.separatorChar + "pom-manip-ext-marker.txt";
 
-    private static final String REPORT_JSON_DEFAULT = File.separatorChar + "manipulation.json";
+    public static final String REPORT_JSON_DEFAULT = "alignmentReport.json";
 
     @ConfigValue( docIndex = "index.html#summary-logging")
     public static final String REPORT_TXT_OUTPUT_FILE = "reportTxtOutputFile";
@@ -221,7 +221,7 @@ public class ManipulationManager
 
                 WildcardMap<ProjectVersionRef> map = (session.getState( RelocationState.class) == null ? new WildcardMap<>() : session.getState( RelocationState.class ).getDependencyRelocations());
                 String report = ProjectComparator.compareProjects( session, jsonReport, map , originalProjects, currentProjects );
-                logger.info( report );
+                logger.info( "{}{}", System.lineSeparator(), report );
 
                 final String reportTxtOutputFile = session.getUserProperties().getProperty( REPORT_TXT_OUTPUT_FILE, "");
                 if ( isNotEmpty( reportTxtOutputFile ) )
--- a/integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
+++ b/integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
@@ -53,6 +53,6 @@ buildLog.eachLine {
 }
 assert executionRoot == true
 
-File json = new File( basedir, 'target/manipulation.json' )
+File json = new File( basedir, 'target/alignmentReport.json' /* + ManipulationManager.REPORT_JSON_DEFAULT*/ )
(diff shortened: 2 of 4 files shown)


integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
```diff
@@ -53,6 +53,6 @@ buildLog.eachLine {
 }
 assert executionRoot == true
 
-File json = new File( basedir, 'target/manipulation.json' )
+File json = new File( basedir, 'target/alignmentReport.json' /* + ManipulationManager.REPORT_JSON_DEFAULT*/ )
```
dwalluck, line 56: Couldn't figure out how to use constant `ManipulationManager.REPORT_JSON_DEFAULT` (not on classpath?).

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -221,7 +221,7 @@ public void scanAndApply( final ManipulationSession session )
 
                 WildcardMap<ProjectVersionRef> map = (session.getState( RelocationState.class) == null ? new WildcardMap<>() : session.getState( RelocationState.class ).getDependencyRelocations());
                 String report = ProjectComparator.compareProjects( session, jsonReport, map , originalProjects, currentProjects );
-                logger.info( report );
+                logger.info( "{}{}", System.lineSeparator(), report );
```
dwalluck, line 224: This adds a newline to the log so that the first line of the output is not misaligned.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -80,7 +80,7 @@
 
     public static final String MARKER_FILE = MARKER_PATH + File.separatorChar + "pom-manip-ext-marker.txt";
 
-    private static final String REPORT_JSON_DEFAULT = File.separatorChar + "manipulation.json";
+    public static final String REPORT_JSON_DEFAULT = "alignmentReport.json";
```
dwalluck, line 83: Made this constant public to reuse (multiple times) from test files.

integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
rnc: They are isolated tests so its likely not on classpath

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
rnc: Nice!

rnc (approved):

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
rnc: Ok

integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
dwalluck: Yeah, if I wanted to reference it, I wasn't sure which POM to add it to. I tried it in the `simple-numeric-directory-path-inverse/pom.xml`, but it didn't seem to work.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
dwalluck: Also, you can see I removed the `File.separatorChar`. I am not sure having it here was intentional, as it adds another `File.separator` later on anyway

integration-test/src/it/simple-numeric-directory-path-inverse/verify.groovy
rnc: They are being run by https://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html , specifically https://maven.apache.org/plugins/maven-invoker-plugin/examples/post-build-script.html

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
rnc: Ok, np.

PR #827, opened 2021-04-26
NCL-6519: Document getRESTAPI() and fix getProjects() documentation

dwalluck:


guide/groovy.md
```diff
@@ -114,14 +114,15 @@ The following API is available:
...
 | [MavenSessionHandler](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/session/MavenSessionHandler.java) getSession() | Return the current session handler |
-| [Properties](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html) getUserProperties() | Get the user properties. |
+| [Properties](https://docs.oracle.com/javase/7/docs/api/java/util/Properties.html) getUserProperties() | Get the user properties |
+| [Translator](https://github.com/release-engineering/pom-manipulation-ext/blob/master/io/src/main/java/org/commonjava/maven/ext/io/rest/Translator.java) getRESTAPI() throws [ManipulationException](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/ManipulationException.java) | Gets a configured VersionTranslator to make REST calls to DA |
```
rnc, line 125: I think this is the wrong place as it wasn't part of the original API so needs moving down to around line 170? (need to check when it was available)

dwalluck: You mean move it to a section without the table? I originally had something there and deleted it. Can't we somehow make a note, like a footnote, in the table? I think it looks better in one place like this.

To summarize: I added `getRESTAPI()` to table and fixed the `getProjects()` return type. I did not find any other missing/incorrect APIs. I also removed some of the periods from the descriptions so that they now consistently don't have periods rather than a mix of both. You could pick one or the other. There's still an inconsistency in the Javadoc itself with no consistent standard used there, but I think the table looks better now. The table/javadoc still mixes verbs, e.g. _get_ vs. _gets_, but seems to use _return_ over _returns_ consistently there. Looking at Java's own docs, they seem to mix also depending on who did the writing.

rnc: There are *multiple* tables depending upon the version of pme. Look at the rendered version https://release-engineering.github.io/pom-manipulation-ext/guide/groovy.html#api .... the caveat with my comment is that I can't remember without checking which version that function call became available in?

dwalluck: > Can't we somehow make a note, like a footnote, in the table?

Or a "Since" column to the table?

guide/groovy.md
```diff
@@ -167,13 +167,23 @@ The following API is available:
...
 
     pme.getBaseDir()
 
 <b>NOTE</b> : Be careful not to use <code>pme.getProperties()</code> or <code>pme.getProject().getProperties()</code> as that actually calls http://docs.groovy-lang.org/latest/html/api/org/codehaus/groovy/runtime/DefaultGroovyMethods.html#getProperties(java.lang.Object)
 
+<table bgcolor="#ffff00">
+<tr>
+<td>
+    <b>NOTE</b> : From version 4.3 the following extra API is available:
+</td>
```
rnc, line 180: I don't think that's right - it's available now just not documented

guide/groovy.md
rnc: Well possibly but that's not how it's been done currently and would be separate IMHO to this work

dwalluck: OK, I moved it. It's in its own section for version 4.3 (that is the next version, right?)

guide/groovy.md
dwalluck: Didn't you need to move it in order to make it accessible? It was `protected` before, so I guess it would be accessible to any script that extended `BaseScript`, but technically wasn't part of the public API.

dwalluck: Looks like: `private` in version 3.8.3, `protected` in 4.0 in `BaseScriptUtils`, but not added to `BaseScript` as `public` until now, which looks to be the file which defines the API.

rnc: So its available from 4.0 ; can you make it that please. If you're not clear on this ping me offline to discuss.

rnc (approved):

Issue #832, opened 2021-05-12
Project version manipulation not working (?!?)


CeddaerrixGE:
I have set the _pom-manipulation-ext_ into _.mvn\extensions.xml_ with 
```
<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
  <extension>
    <groupId>org.commonjava.maven.ext</groupId>
    <artifactId>pom-manipulation-ext</artifactId>
    <version>4.3</version>
  </extension>
</extensions>
```

From the root of the multi-project project,  have then run the following:
`mvn -DversionSuffix=Some-Suffix clean install -e -B -f pom.xml -pl :project-common -am -P<some profiles> -s settings-maven.xml`

In the target folder, the package remains: _project-common-3.4.0-SNAPSHOT.jar_ and not _project-common-3.4.0-SNAPSHOT-Some-Suffix.jar_

Should not it behave as if having the _pom-manipulation-ext-4.3.jar_ was in the _%MAVEN_HOME\lib\ext_ directory?

rnc:
I don't think we have tested it as a core extension as opposed to a lib/ext. Is this use-case something you would need?

You may find it easier to use it directly as a CLI tool (which is the primary mechanism we are currently using it with) ; see https://release-engineering.github.io/pom-manipulation-ext/#installation-as-cli-tool

CeddaerrixGE:
> I don't think we have tested it as a core extension as opposed to a lib/ext. Is this use-case something you would need?
> 
> You may find it easier to use it directly as a CLI tool (which is the primary mechanism we are currently using it with) ; see https://release-engineering.github.io/pom-manipulation-ext/#installation-as-cli-tool

I have made it work with 'lib/ext'...
Still, I would have found it more elegant and convenient to have it configured into _.mvn\extensions.xml_ on the project SCM repo, thus not to have to find out where is M2_HOME on a Jenkins agent to have the jar copied into 'lib/ext' for pipelines using that extension

rnc:
[...]

PR #833, opened 2021-05-21
Modify shading so library is picked up as core extension

rnc:
For #832

While this PR resolves _detection_ of the extensions there appears to be further classloader issues in that Maven Model classes are loaded within  `ClassRealm[plexus.core` versus `ClassRealm[coreExtension` which causes issues

### DIFF
--- a/ext/pom.xml
+++ b/ext/pom.xml
@@ -89,12 +89,6 @@
     </dependency>
 
     <dependency>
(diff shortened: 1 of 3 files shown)


rnc:
I'm merging this as it doesn't break `lib/ext`, fixes problems with shading (although it does then demonstrate the problem with classloaders)

PR #838, opened 2021-06-09
NCLSUP365 Scan all pom files even those with variables in groupId/art…

rnc:
…ifactId

### DIFF
--- a/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
+++ b/io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
@@ -15,6 +15,23 @@
  */
 package org.commonjava.maven.ext.io;
 
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Set;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.reflect.FieldUtils;
 import org.apache.maven.model.Model;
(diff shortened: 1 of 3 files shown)


io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
```diff
@@ -64,6 +67,9 @@
 @Singleton
 public class PomIO
 {
+    // TODO: Remove this if no side affects reported in 2022.
+    public static final String PARSE_POM_TEMPLATES = "parsePomTemplates";
```
dwalluck, line 71: Should this be `public` or `private`? It currently seems only used within the class.

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
```diff
@@ -73,8 +79,23 @@
...
 
+    @Inject
+    public PomIO(MavenSessionHandler handler )
+    {
+        parsePomTemplates = Boolean.parseBoolean(
+                        handler.getUserProperties().getProperty( PARSE_POM_TEMPLATES, "true" ) );
+    }
+
+    // Test use only.
+    public PomIO()
```
dwalluck, line 94: I am not sure I see this constructor used anywhere currently. If it's used within the same package, it could be made package-private.

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
rnc: No public - its used in a test

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
rnc: Vast amounts of tests use it so seemed simpler to create extra ctor.

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
dwalluck: Oh, there using the default public constructor? I think some version of Java is getting rid of those and they have to be made explicit.

io/src/main/java/org/commonjava/maven/ext/io/PomIO.java
dwalluck: Oh, they were using the default constructor? I think the some version of Java is getting rid of those and they have to be written explicitly.

rnc: But for now I think its acceptable?

dwalluck: Yes, it was just for your information, and we have an explicit constructor now.

I don't know if it will ever be removed, but there are some bugs that allow a warning to be enabled as of Java 16, and specifically for public APIs.

https://bugs.openjdk.java.net/browse/JDK-8071961
https://bugs.openjdk.java.net/browse/JDK-8250795
https://bugs.openjdk.java.net/browse/JDK-8252348

rnc: Thanks.

PR #839, opened 2021-06-11
NCL-6611 NCL-6528 Switch to new lookup/maven DA API

rnc:


common/src/main/java/org/commonjava/maven/ext/common/util/GAVUtils.java
```diff
@@ -19,17 +19,18 @@
...
 /**
  * Commonly used manipulations from project profiles.
  */
 @UtilityClass
 public final class GAVUtils
 {
-    public static List<GAV> generateGAVs( List<ProjectVersionRef> dep) {
-        ArrayList<GAV> result = new ArrayList<>();
+    public static Set<GAV> generateGAVs( List<ProjectVersionRef> dep) {
+        Set<GAV> result = new HashSet<>();
```
dwalluck, line 33: Maybe order does not matter here, but the original `List` is ordered, so wondered if we'd want to maintain the order and use `LinkedHashSet` instead.

io/src/test/java/org/commonjava/maven/ext/io/rest/handler/GAVSchema.java
```diff
@@ -16,47 +16,22 @@
...
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+@ToString
 public class GAVSchema
 {
     @JsonInclude(value = JsonInclude.Include.NON_EMPTY)
-    public String versionSuffix;
+    public List<Map<String, Object>> artifacts;
```
dwalluck, line 28: Any reason why these are all public fields instead of using the getter/setter convention?

rnc: Nothing really...its just test code so kept it easy to access

common/src/main/java/org/commonjava/maven/ext/common/util/GAVUtils.java
rnc: I don't believe order matters , I think it's just too pass a collection (of some form) through to DA

rnc: Its for https://github.com/project-ncl/dependency-analysis/blob/master/reports-model/src/main/java/org/jboss/da/lookup/model/MavenLookupRequest.java#L20

PR #842, opened 2021-06-15
Documentation improvements. Remove deprecated annotation. Add note for Translator changes.

rnc:


stylesheets/stylesheet.css
```diff
@@ -219,7 +219,7 @@ Full-Width Styles
 
 .inner {
   position: relative;
-  max-width: 640px;
+  max-width: 75%;
```
rnc, line 222: @dwalluck I think we should apply this to GME as well. Can you do this please?

dwalluck: Sure. I suppose that fixed width makes more sense on smaller devices. You can see it used elsewhere, particularly for other media types. For a large screen, I am not sure why there is such a strict limit.

dwalluck: https://github.com/project-ncl/gradle-manipulator/pull/296

rnc: Thanks!

rnc:
> Pending the official API name change, of course. Do we the ticket numbers on the first line?

Sorry not sure what you mean re ticket numbers - I have referenced the JIRAs in the commits.

PR #846, opened 2021-06-17
NCL6600 Code improvements

rnc:


io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -58,14 +58,15 @@
     {
     };
 
-    public enum ENDPOINT {
+    public enum Endpoint
+    {
         LOOKUP_GAVS ("lookup/maven"),
         LOOkUP_LATEST ("lookup/maven/latest");
```
dwalluck, line 64: Wrong case here: k --> K

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
```diff
@@ -58,14 +58,15 @@
...
 
         @Getter
         private final String endpoint;
 
-        ENDPOINT( String s )
```
dwalluck, line 68: Should we override `toString()` to return `getEndpoint()`? I sort of expected this so that I could build a URL using `"/" + LOOKUP_GAVS`. What do you think?

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Great spot! Thanks!

io/src/main/java/org/commonjava/maven/ext/io/rest/DefaultTranslator.java
rnc: Thats a good idea. Could probably remove the `@Getter` maybe?

PR #849, opened 2021-06-23
Check if /proc/1/cgroup is readable before trying to open

dwalluck:


cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
```diff
@@ -454,26 +455,31 @@ private Settings parseSettings( File settings ) throws ComponentLookupException,
...
         {
             result = System.getenv().containsKey( "container" );
         }
+
         return result;
     }
 
     // Split into separate method for testing only.
-    private String getCGroups()
+    private static String getCGroups()
```
rnc, line 482: Will the byteman test still work now you have changed it to static?

dwalluck: Yes, the `CliLoggingTest` passed with no changes to that BM annotation after making this change, but I could revert it if you want.

My reasoning was that the method could be static and I was a bit confused since the `getCGroups` method itself was not static, but the fake cgroup results that we return from the tests for this method are static (if I am making sense---they might not be related at all).

dwalluck:
Well, codecov/patch is failing because we don't cover the failure cases with a test, I suppose. Should a new test method be added?

cli/src/main/java/org/commonjava/maven/ext/cli/Cli.java
rnc: Thats fine, I couldn't remember off top of my head whether the byteman would work with the change.

rnc:
Whats worse, its again picking up MavenLocationExpander and saying the coverage on that has changed. When you didn't touch that file. Arg!

rnc:
Can you rebase on top of #853 please?

dwalluck:
I added two extra test methods and it now seems to break byteman.

dwalluck:
> [@dwalluck ](/dwalluck)
> ` [Add test for unreadable cgroup file](/release-engineering/pom-manipulation-ext/pull/849/commits/8da77f64f9c5562fbffafafba5901fa4052b2ea7) `
> ` [8da77f6](/release-engineering/pom-manipulation-ext/pull/849/commits/8da77f64f9c5562fbffafafba5901fa4052b2ea7)`

I am not sure if the removal of this test was intentional, but I put it back and it passes on Windows right now, but not Linux. Unfortunately, the console output shows almost nothing.

Also, I wonder if the logger reset should be moved to `@After` so that it runs after each test. It is not clear to me why certain classes which use the `-l` flag don't have it.

dwalluck:
> It passes on Windows right now, but not Linux. Unfortunately, the console output shows almost nothing.

I managed to run it locally:

```
java.io.UncheckedIOException: java.io.IOException: Is a directory
	at java.base/java.nio.file.FileChannelLinesSpliterator.readLine(FileChannelLinesSpliterator.java:176)
	at java.base/java.nio.file.FileChannelLinesSpliterator.tryAdvance(FileChannelLinesSpliterator.java:104)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:230)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:196)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.anyMatch(ReferencePipeline.java:528)
	at org.commonjava.maven.ext.cli.Cli.runningInContainer(Cli.java:465)
	at org.commonjava.maven.ext.cli.Cli.call(Cli.java:172)
	at org.commonjava.maven.ext.cli.Cli.call(Cli.java:83)
```

PR #852, opened 2021-06-23
Fix some javadoc warnings

dwalluck:


rnc:
@dwalluck Codecov is strange - how can adding docs lead to a **drop** in code coverage!!

PR #860, opened 2021-06-29
Fix log messages

dwalluck:


core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -158,23 +159,31 @@ void applyGroovyScript( List<Project> projects, Project project, File groovyScri
         {
             try
             {
-                logger.debug ( "Failure when injecting into script {} ", FileUtils.readFileToString( groovyScript, StandardCharsets.UTF_8 ), e );
+                if ( logger.isDebugEnabled() )
+                {
+                    logger.debug( "Failure when injecting into script {}",
+                            FileUtils.readFileToString( groovyScript, StandardCharsets.UTF_8 ), e );
```
rnc, line 165: I wonder if these actually should be `logger.error` given we are in a failure case (which would make the if block redundant)

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
```diff
@@ -158,23 +159,31 @@ void applyGroovyScript( List<Project> projects, Project project, File groovyScri
...
         }
         catch (CompilationFailedException e)
         {
             try
             {
-                logger.debug ( "Failure when parsing script {} ", FileUtils.readFileToString( groovyScript, StandardCharsets.UTF_8 ), e );
+                if ( logger.isDebugEnabled() )
+                {
+                    logger.debug( "Failure when parsing script {}",
+                            FileUtils.readFileToString( groovyScript, StandardCharsets.UTF_8 ), e );
```
rnc, line 181: As per above, I wonder if these actually should be `logger.error` given we are in a failure case (which would make the if block redundant)

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -344,17 +344,23 @@ private void apply( final Project project, final Model model, final Map<Artifact
                                 }
                                 else
                                 {
-                                    logger.warn( "Parent reference {} replacement: {} of original version: {} violates the strict version-alignment rule!",
-                                                 ga( project.getModelParent() ), newValue, oldValue );
+                                    if ( logger.isWarnEnabled() )
```
rnc, line 347: Probably redundant as we also have warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyRemovalManipulator.java
```diff
@@ -99,7 +99,10 @@ private boolean apply( final Project project, final Model model ) throws Manipul
     {
         final DependencyRemovalState state = session.getState(DependencyRemovalState.class);
 
-        logger.info("Applying Dependency changes to: " + ga(project));
+        if ( logger.isInfoEnabled() )
```
rnc, line 102: Probably redundant as we also have info level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/ProfileInjectionManipulator.java
```diff
@@ -90,10 +91,14 @@ public void init( final ManipulationSession session )
             final List<Profile> remoteProfiles = modelBuilder.resolveRawModel( p ).getProfiles();
 
             projects.stream().filter( Project::isInheritanceRoot ).forEach( project -> {
-                    logger.info( "Applying changes to: {} ", ga( project ) );
-                    project.updateProfiles( remoteProfiles );
-                    changed.add( project );
-                } );
+                if ( logger.isInfoEnabled() )
```
rnc, line 94: Probably redundant as we also have info/warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/ProfileRemovalManipulator.java
```diff
@@ -76,8 +76,10 @@ public void init( final ManipulationSession session )
 
         for ( final Project project : projects )
         {
-            final String ga = ga( project );
-            logger.info( "Applying changes to: " + ga );
+            if ( logger.isInfoEnabled() )
```
rnc, line 79: Probably redundant as we also have info/warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/PropertyManipulator.java
```diff
@@ -90,7 +90,10 @@ public void init( final ManipulationSession session )
                 // Only inject the new properties at the top level.
                 if ( project.isInheritanceRoot() )
                 {
-                    logger.info( "Applying property changes to: {} with {}", ga( project ), overrides );
+                    if ( logger.isInfoEnabled() )
```
rnc, line 93: Probably redundant as we also have info/warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/PropertyManipulator.java
```diff
@@ -103,13 +106,17 @@ public void init( final ManipulationSession session )
...
 
-                    if (!keyClone.isEmpty())
+                    if ( !keyClone.isEmpty() )
                     {
                         final Iterator<String> keys = keyClone.iterator();
                         while (keys.hasNext())
                         {
                             final String matchingKey = keys.next();
-                            logger.info( "Overwriting property ({} in: {} with value {}", matchingKey, ga( project ), overrides.get( matchingKey ));
+                            if ( logger.isInfoEnabled() )
```
rnc, line 115: Probably redundant as we also have info/warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/RepositoryInjectionManipulator.java
```diff
@@ -93,8 +93,11 @@ public void init( final ManipulationSession session )
 
         for ( final Project project : projects )
         {
-            final String ga = ga( project );
-            logger.info( "Applying changes to: " + ga );
+            if ( logger.isInfoEnabled() )
```
rnc, line 96: Probably redundant as we also have info/warn level enabled

common/src/main/java/org/commonjava/maven/ext/common/util/WildcardMap.java
```diff
@@ -127,7 +127,11 @@ public void put(ProjectRef key, T value)
             // Erase any previous mappings.
             if (!vMap.isEmpty())
             {
-                logger.warn ("Emptying map with keys " + vMap.keySet() + " as replacing with wildcard mapping " + key);
+                if ( logger.isWarnEnabled() )
```
rnc, line 130: Probably redundant as we also have info/warn level enabled

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
dwalluck: The guard is there in case there is an expensive method call (e.g., not a simple getter) because we pay the price of the method call whether or not the log is enabled (printed).

The case of reading a file seems particularly expensive. Inside a `catch` block may not be as bad since this is an abnormal code path to begin with.

And it may be slightly worse in debug level since it is not enabled by default, but if we also have a quiet mode, then we'd pay the price there as well.

As an aside, I assumed it was intentional, but is the exception `e` passed as an argument there just meant to print `e.toString()` or do we mean to pass the `Throwable` itself as an additional argument (which also shows the stack trace?)

rnc: I think these should be logger.error (without the guard)

dwalluck: I will change it to error level.

Keep in mind that if we remove the guard, then even if we set the logger via `setLevel( Level.OFF )`, the file still gets read, which doesn't seem right. But, I will remove that if you want.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: The issue with the `ga()` method call is a that it's calling `String.format()` inside a logger call, so we lose the point of using the `{}` placeholders.

A simple fix is to remove the guard and replace `String.format("%s:%s")` with `"logger.info("{}:{}", project.getGroupId(), project.getArtifactId())`.

core/src/main/java/org/commonjava/maven/ext/core/impl/BaseGroovyManipulator.java
rnc: I think at that point we need to supply as much information as possible as to what caused the error fir the user to understand. As the build will fail, reading a file doesn't matter in my opinion

dwalluck: OK, I've removed the conditional on these and changed it to error level.

I also changed to pass e as an argument (not as a string) as this should provide more information about the error and may be what was intended (not sure), but I think it is better that way as it will have the stack trace.

PR #866, opened 2021-08-02
NCLSUP-377: Guard against null plugin version

dwalluck:
We have a `Map<ProjectVersionRef, Plugin>` where `plugin.getVersion()` is `null`. As seen below, the `plugin` value has no version specified in the POM. However, the key `projectVersionRef` has a version specified (`org.apache.maven.plugins:maven-surefire-plugin:2.12.4`).

```
            <pluginManagement>
                <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-surefire-plugin</artifactId>
                    <configuration>
                        <forkMode>once</forkMode>
                        <argLine>-Djava.awt.headless=true ${surefire.memory.settings}</argLine>
                        <runOrder>alphabetical</runOrder>
                    </configuration>
                </plugin>
            </pluginMangement>
```

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/impl/ProjectVersioningManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/impl/ProjectVersioningManipulator.java
@@ -43,6 +43,7 @@ import java.util.Map;
 import java.util.Set;
 
 import static org.apache.commons.lang.StringUtils.isEmpty;
+import static org.apache.commons.lang.StringUtils.startsWith;
 
 /**
  * {@link Manipulator} implementation that can modify a project's version with either static or calculated, incremental version qualifier. Snapshot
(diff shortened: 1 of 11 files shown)


dwalluck:
Is this all that is needed or is something else wrong? This same pattern of `startsWith` without any `null` check is used in many other places in the code, and we didn't git a `null` until now. Also, what is the relationship of the plugin version of `null` to the dependency version of `"2.12.4"` here?

rnc:
I think that while the `getResolvedPlugins` / `getResolvedManagedPlugins` should have handled this there are some corner cases being exposed now especially with the managed plugins. I don't think we have any integration tests covering this (I suspect adding a similar block to what Keycloak has in their pluginMangement) would replicate it the problem. There may also be an issue on line 395 in PluginManipulator but I haven't replicated that. Before this can be merged it needs a test - I suggest adapting an existing test to add a parent pom e.g. jboss-parent and a versionless plugin.

rnc:
I have a potentially simpler fix within `resolvePlugins` ; whether they should be combined is another question.

dwalluck:
> I have a potentially simpler fix within `resolvePlugins` ; whether they should be combined is another question.

That would make this change unnecessary? In any case, I was going to at a testcase to this pull request.

rnc:
Lets discuss offline.

rnc:
Looks good, I would just like a test as per our meeting

core/src/main/java/org/commonjava/maven/ext/core/impl/ProjectVersioningManipulator.java
```diff
@@ -283,8 +284,8 @@ else if ( !changed && project.isInheritanceRoot() )
 
                         if ( newVersion != null && d.getVersion() != null )
```
dwalluck, line 285: Do we need the check `d.getVersion() != null` here?

rnc: It is passed into updateProperties so I'd suggest keeping it.

rnc (approved):

PR #868, opened 2021-08-06
NCL-6697 Handle artifactId/groupId being property expressions. Add pluginRelocation functionality

### LINKED ISSUE #867: dependencyRelocation ignores properties used for groupId and artifactId and does not work for plugins
We're in the process of updating our Quarkus Quickstarts to use some similar property mechanism to Quarkus[1] which allows users to override the BOM to something else.

This[2] was the result of running PME across one of the Quarkus examples in kogito-examples and it successfully replaces the imported BOM in the dependencyManagement but doesn't trace it back and replace the property instead. It also ignores the plugin (using the property <quarkus-plugin.version>).

This is what I tried:

-DdependencyRelocations.io.quarkus:quarkus-bom@com.redhat:some-artifact=1.2.3.redhat-00001 -DdependencyRelocations.io.quarkus:quarkus-maven-plugin@com.redhat:some-plugin=123.redhat-00001

My expectation is that it updates the properties instead, like it does for the `<quarkus.platform.version>` in the example below. It should also be able to take care of a plugin specified in a similar manner too.

[1] https://github.com/quarkusio/quarkus-quickstarts/blob/main/kogito-quickstart/pom.xml#L12-L15
[2] 
```
    diff --git a/decisiontable-quarkus-example/pom.xml b/decisiontable-quarkus-example/pom.xml
    index 0dc2f849..3f3cb134 100644
    --- a/decisiontable-quarkus-example/pom.xml
    +++ b/decisiontable-quarkus-example/pom.xml
    @@ -1,6 +1,5 @@
     <?xml version="1.0" encoding="UTF-8"?>
    -<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    -    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
    +<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
       <modelVersion>4.0.0</modelVersion>
       <parent>
         <groupId>org.kie.kogito.examples</groupId>
    @@ -13,13 +12,13 @@
         <quarkus-plugin.version>2.1.0.Final</quarkus-plugin.version>
         <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
         <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
    -    <quarkus.platform.version>2.1.0.Final</quarkus.platform.version>
    +    <quarkus.platform.version>1.2.3.redhat-00001</quarkus.platform.version>
       </properties>
       <dependencyManagement>
         <dependencies>
           <dependency>
    -        <groupId>${quarkus.platform.group-id}</groupId>
    -        <artifactId>${quarkus.platform.artifact-id}</artifactId>
    +        <groupId>com.redhat</groupId>
    +        <artifactId>some-artifact</artifactId>
             <version>${quarkus.platform.version}</version>
             <type>pom</type>
             <scope>import</scope>
```

rnc:
Fixes #867

### DIFF
--- a/common/src/main/java/org/commonjava/maven/ext/common/util/WildcardMap.java
+++ b/common/src/main/java/org/commonjava/maven/ext/common/util/WildcardMap.java
@@ -21,7 +21,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.LinkedHashMap;
-import java.util.TreeMap;
+import java.util.Map;
 
 /**
  * Custom limited map implementation that handles the following format:
@@ -43,7 +43,7 @@ public class WildcardMap<T>
      * </p>
      * artifactId may be a wildcard '*'.
      */
-    private final TreeMap<String, LinkedHashMap<String,T>> map = new TreeMap<>();
+    private final Map<String, Map<String,T>> map = new LinkedHashMap<>();
 
     /**
      * Size implementation
@@ -84,7 +84,7 @@ public class WildcardMap<T>
     {
         boolean result;
 
-        LinkedHashMap<String, T> vMap = map.get( groupId);
+        Map<String, T> vMap = map.get( groupId);
 
         if ( vMap == null || vMap.isEmpty())
         {
@@ -115,7 +115,7 @@ public class WildcardMap<T>
         String groupId = key.getGroupId();
         String artifactId = key.getArtifactId();
 
-        LinkedHashMap<String,T> vMap = map.get(groupId);
+        Map<String,T> vMap = map.get(groupId);
         if ( vMap == null)
         {
             vMap = new LinkedHashMap<>();
@@ -188,7 +188,7 @@ public class WildcardMap<T>
     {
         T result = null;
 
-        LinkedHashMap<String, T> value = map.get(groupId);
+        Map<String, T> value = map.get(groupId);
         if (value != null)
         {
             if ( value.get(WILDCARD) != null)
--- a/core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
@@ -28,6 +28,7 @@ import org.codehaus.plexus.util.xml.Xpp3DomBuilder;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.commonjava.maven.atlas.ident.ref.ArtifactRef;
 import org.commonjava.maven.atlas.ident.ref.ProjectVersionRef;
+import org.commonjava.maven.atlas.ident.ref.SimpleProjectVersionRef;
 import org.commonjava.maven.ext.common.ManipulationException;
 import org.commonjava.maven.ext.common.model.Project;
 import org.commonjava.maven.ext.common.model.SimpleScopedArtifactRef;
@@ -65,6 +66,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
+
+import static org.apache.commons.lang.StringUtils.isEmpty;
 
 /**
  * {@link Manipulator} implementation that can relocation specified groupIds. It will also handle version changes by
@@ -136,42 +139,43 @@ public class RelocationManipulator
     {
         boolean result = false;
         final RelocationState state = session.getState( RelocationState.class );
-        final WildcardMap<ProjectVersionRef> relocations = state.getDependencyRelocations();
-        // TODO: Implement getPluginRelocations and use that instead below
+        final WildcardMap<ProjectVersionRef> dependencyRelocations = state.getDependencyRelocations();
+        final WildcardMap<ProjectVersionRef> pluginRelocations = state.getPluginRelocations();
 
-        logger.debug( "Applying relocation changes ({}) to: {}:{}", relocations, project.getGroupId(),
-                project.getArtifactId() );
+        logger.debug( "Applying relocation changes for dependencies ({}) and for plugins ({}) to: {}:{}",
+                      dependencyRelocations, pluginRelocations,
+                      project.getGroupId(), project.getArtifactId() );
 
         DependencyManagement dependencyManagement = model.getDependencyManagement();
         if ( dependencyManagement != null )
         {
-            result = updateDependencies( project, relocations, project.getResolvedManagedDependencies( session ) );
+            result = updateDependencies( project, dependencyRelocations, project.getResolvedManagedDependencies( session ) );
         }
-        result |= updateDependencies( project, relocations, project.getAllResolvedDependencies( session ) );
+        result |= updateDependencies( project, dependencyRelocations, project.getAllResolvedDependencies( session ) );
 
         for ( final Profile profile : ProfileUtils.getProfiles( session, model) )
         {
             dependencyManagement = profile.getDependencyManagement();
             if ( dependencyManagement != null )
             {
-                result |= updateDependencies( project, relocations, project.getResolvedProfileManagedDependencies( session ).get( profile ) );
+                result |= updateDependencies( project, dependencyRelocations, project.getResolvedProfileManagedDependencies( session ).get( profile ) );
             }
-            result |= updateDependencies( project, relocations, project.getAllResolvedProfileDependencies( session ).get( profile ) );
+            result |= updateDependencies( project, dependencyRelocations, project.getAllResolvedProfileDependencies( session ).get( profile ) );
 
         }
 
-        result |= updatePluginConfigurations( relocations, project, project.getResolvedManagedPlugins( session ) );
+        result |= updatePlugins( pluginRelocations, dependencyRelocations, project, project.getResolvedManagedPlugins( session ) );
 
-        result |= updatePluginConfigurations( relocations, project, project.getAllResolvedPlugins( session ) );
+        result |= updatePlugins( pluginRelocations, dependencyRelocations, project, project.getAllResolvedPlugins( session ) );
 
         for ( Profile profile : project.getAllResolvedProfilePlugins( session ).keySet() )
         {
-            result |= updatePluginConfigurations( relocations, project, project.getAllResolvedProfilePlugins( session ).get( profile ) );
+            result |= updatePlugins( pluginRelocations, dependencyRelocations, project, project.getAllResolvedProfilePlugins( session ).get( profile ) );
         }
 
         for ( Profile profile : project.getResolvedProfileManagedPlugins( session ).keySet() )
         {
-            result |= updatePluginConfigurations( relocations, project, project.getResolvedProfileManagedPlugins( session ).get( profile ) );
+            result |= updatePlugins( pluginRelocations, dependencyRelocations, project, project.getResolvedProfileManagedPlugins( session ).get( profile ) );
         }
 
         return result;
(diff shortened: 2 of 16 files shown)


integration-test/src/it/relocation-group/verify.groovy
```diff
@@ -46,3 +46,13 @@ assert profile1.size() != 0
 dependency = pom.dependencyManagement.dependencies.dependency.find { it.artifactId.text() == "junit-dep" }
 assert dependency != null
 assert dependency.version.text() == "4.1"
+
+
+def passed = false
+pom.properties.each {
+    if ( it.text().contains ("slf4j") )
+    {
+        passed = true
```
dwalluck, line 55: This can't break, so once true, always true. Maybe we just want a `for` loop instead? Or `find`? Several ways to do it.

rnc: Indeed ; it was a copy and paste from how other tests did it.

rnc: Given that there are so many of this pattern I'm think it can be left as is ; any changes should be done en-mass in a separate PR IMHO.

integration-test/src/it/relocation-plugin/pom.xml
```diff
@@ -59,7 +60,7 @@
     <plugins>
       <plugin>
         <groupId>com.soebes.maven.plugins</groupId>
-        <artifactId>iterator-maven-plugin</artifactId>
+        <artifactId>repeater-maven-plugin</artifactId>
```
dwalluck, line 63: Why this change. It looks like iterator-maven-plugin does exist, but no repeater-maven-plugin exists: https://search.maven.org/search?q=g:com.soebes.maven.plugins.

integration-test/src/it/relocation-group/verify.groovy
```diff
@@ -46,3 +46,13 @@ assert profile1.size() != 0
 dependency = pom.dependencyManagement.dependencies.dependency.find { it.artifactId.text() == "junit-dep" }
 assert dependency != null
 assert dependency.version.text() == "4.1"
+
+
+def passed = false
+pom.properties.each {
+    if ( it.text().contains ("slf4j") )
```
dwalluck, line 53: To make it a little clearer, are we looking for "org.slf4j" specifically? I would make it more specific only because both the groupId and artifactId contain "slf4j" and to me it's just a tiny bit clearer which part we are checking for.

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -308,6 +326,36 @@ private boolean updatePlugins( final WildcardMap<ProjectVersionRef> relocations,
...
+     * @param originalValue the original property value e.g. the value of {@code artifactId}
+     * @param originalRelocation the complete original relocation
+     * @param relocation the portion of the relocation that is currently being processed
+     * @param c a Consumer function that will update the corresponding Dependency/Plugin if its not a property.
+     * @throws ManipulationException if an error occurs.
+     */
+    private void updateString( Project project, String originalValue, ProjectVersionRef originalRelocation, String relocation,
+                               Consumer<String> c ) throws ManipulationException
+    {
+        if ( StringUtils.contains( originalValue, "$" ))
```
dwalluck, line 343: Only a style comment, but in other code recently I used a static import for any `StringUtils` methods.

integration-test/src/it/relocation-plugin/pom.xml
rnc: Er....because we are relocating as a test?

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -230,37 +236,46 @@ private boolean updateDependencies( Project project, WildcardMap<ProjectVersionR
...
-            {
-                throw new ManipulationException( "Relocations with multiple embedded version properties not supported" );
-            }
-            logger.debug( "Updating relocation for {} with property {} to new version {}",
-                          dependency == null ? versionNode : dependency, property, relocation.getVersionString() );
-            PropertiesUtils.updateProperties( session, project, true, property, relocation.getVersionString() );
-        }
-        else
+        // Handles plugins themselves
+        for ( int i = 0 ; i < pluginRelocations.size(); i++ )
```
dwalluck, line 243: This loop (and the similar one below) doesn't look right to me. From what I can see, we're never accessing this value and we're not really iterating over this map, but I different one.

integration-test/src/it/relocation-group/verify.groovy
rnc: It to verify that `slf4j` gets substituted into `<slf4j>slf10</slf4j>`

core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java
```diff
@@ -230,37 +236,46 @@ private boolean updateDependencies( Project project, WildcardMap<ProjectVersionR
...
         {
-            if ( dependency == null )
-            {
-                versionNode.setTextContent( relocation.getVersionString() );
-            }
-            else
+            Iterator<ProjectVersionRef> it = pluginMap.keySet().iterator();
+            while ( it.hasNext() )
```
dwalluck, line 246: If iterating over map, should avoid `get` and use pattern:

 ```
for ( Entry<ProjectVersionRef, Plugin> entry: pluginMap.entrySet() )
 {
    final ProjectVersionRef pvr = entry.getKey();
    final Plugin plugin = entry.getValue();
}
```

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 4 above)
Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:343]: Pros/cons ... I like to make it obvious where its coming from.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java [hunk 7]
```diff
@@ -272,17 +287,19 @@ private boolean updatePlugins( final WildcardMap<ProjectVersionRef> relocations,
                 dependency.setGroupId( pluginReference.groupIdNode.getTextContent() );
                 dependency.setArtifactId( pluginReference.artifactIdNode.getTextContent() );
 
-                final boolean isUpdate = relocations.containsKey( dependency );
+                final boolean isUpdate = dependencyRelocations.containsKey( dependency );
```
dwalluck, line 290: Is `containsKey` necessary, or will it get return `null` just in this case? Then, we could combine this line and 294 as

```
final ProjectVersionRef relocation = dependencyRelocations.get( dependency );

if ( relocation != null )
```
```

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 5 above)
Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:243]: Same pattern as used in dependencies so should be ok.

File: integration-test/src/it/relocation-plugin/pom.xml (hunk 2 above)
Comment (dwalluck) [reply integration-test/src/it/relocation-plugin/pom.xml:63]: Oops, I missed there were two different relocations in this test.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 7 above)
Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:290]: I've not changed this apart from renaming the variable so this your original code.

File: integration-test/src/it/relocation-group/verify.groovy (hunk 3 above)
Comment (dwalluck) [reply integration-test/src/it/relocation-group/verify.groovy:53]: I thought it was `slf10 --> org.slf4j`? That's why I wasn't 100% clear there based on the test (for me).

Comment (rnc) [reply integration-test/src/it/relocation-group/verify.groovy:53]: Yeah it is, I'll change it.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 5 above)
Comment (dwalluck) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:246]: I know the code has been there a long time, but it's not making sense to me. There's a double loop and the outer loop never appears to be used. The problem is `i` is never used. Is `relocations` ever used "sequentially" so that we're not looping `relocations.size()` extra times or something like this?

It has

```
        // If we do a single pass over the dependencies that will handle the relocations *but* it will not handle
        // where one relocation alters the dependency and a subsequent relocation alters it again. For instance, the
        // first might wildcard alter the groupId and the second, more specifically alters one with the artifactId
        for ( int i = 0 ; i < relocations.size(); i++ )
        {
            Iterator<ArtifactRef> it = dependencies.keySet().iterator();
            while ( it.hasNext() )
            {
                final ArtifactRef pvr = it.next();
                if ( relocations.containsKey( pvr.asProjectRef() ) )
                {
                    ProjectVersionRef relocation = relocations.get( pvr.asProjectRef() );

```



It's not as if `relocation = relocations.get( i )` as in a normal loop. Instead, `relocation =  relocations.get( pvr.asProjectRef() )` where `pvr` comes from the loop over `dependencies`, not `relocations`.

Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:246]: Right but my point is you didn't change it in your original pr so why highlight it now? Currently I am not handling the iteration removal for plugins same as dependencies but reused the same pattern of code as I presumed its working.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 4 above)
Comment (dwalluck) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:346]: That's fine either way, but maybe we should have enforced a single style across the project.

Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:346]: True we don't really specify that :-( Happy to discuss offline ; pros/cons to both approaches.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 7 above)
Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java:293]: But I do agree with your assessment and will change it.

File: core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java (hunk 5 above)
Comment (rnc) [reply core/src/main/java/org/commonjava/maven/ext/core/impl/RelocationManipulator.java]: I've added a test for multiple plugin relocations and double checked the code.

File: common/src/main/java/org/commonjava/maven/ext/common/util/WildcardMap.java [hunk 8]
```diff
@@ -43,7 +42,7 @@
      * </p>
      * artifactId may be a wildcard '*'.
      */
-    private final TreeMap<String, LinkedHashMap<String,T>> map = new TreeMap<>();
+    private final LinkedHashMap<String, LinkedHashMap<String,T>> map = new LinkedHashMap<>();
```
dwalluck, line 45: Want `Map` as the types on the left?

rnc: Could do

PR #871, opened 2021-08-08
Fix map iteration

dwalluck:
Avoid (possibly multiple) `get()` calls when iterating over maps.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/impl/SuffixManipulator.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/impl/SuffixManipulator.java
@@ -37,6 +37,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -63,8 +64,6 @@ public class SuffixManipulator
         this.session = session;
         session.setState( new SuffixState( session.getUserProperties() ) );
     }
-
-
 
     /**
      * Apply the property changes to the list of {@link Project}'s given.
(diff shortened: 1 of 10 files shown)


rnc:
Good idea

rnc:
While I'm fine with this and it looks correct to me, I guess _techinically_ using `map.forEach((k,v) -> { ....` would be slightly more efficient than a `for` and `Map.Entry` although _not_ applicable to every case - as break/continue/exceptions could cause problems with the lambda statements.

dwalluck:
> While I'm fine with this and it looks correct to me, I guess _techinically_ using `map.forEach((k,v) -> { ....` would be slightly more efficient than a `for` and `Map.Entry` although _not_ applicable to every case - as break/continue/exceptions could cause problems with the lambda statements.

I don't see how. I believe the `Map.forEach()` implementation is basically equivalent to `Map.entrySet().forEach()`, so the performance should be nearly identical  to `for-each-entrySet()`, but with some slight overhead.

Also, in addition to the drawbacks of lambdas that you point out, lambdas can  sometimes get unwieldly when they're more than a few lines or nested.

In the current code, there was only a single instance of `Map.forEach()` that I recall, and I changed it to match the pattern of the others. It currently keeps this same pattern whether using `keySet()`, `values()`, or `entrySet()`.

core/src/main/java/org/commonjava/maven/ext/core/impl/SuffixManipulator.java
```diff
@@ -136,17 +135,20 @@ private void processPlugins( Pattern suffixStripPattern, Project project,
         {
             if ( plugins != null )
             {
-                plugins.keySet().forEach( a -> {
```
dwalluck, line 139: This was the only existing `.forEach()`.

rnc (approved):

PR #875, opened 2021-08-12
NCL-6658 Update to JDOM2 and resolve CVE

rnc:


core/src/main/java/org/commonjava/maven/ext/core/state/RelocationState.java
```diff
@@ -57,7 +57,7 @@
     @ConfigValue( docIndex = "dep-manip.html#dependency-relocations" )
     public static final String DEPENDENCY_RELOCATIONS = "dependencyRelocations.";
 
-    @ConfigValue( docIndex = "plugin-manip.html#dependency-relocations" )
+    @ConfigValue( docIndex = "plugin-manip.html#plugin-relocations" )
```
dwalluck, line 60: I assume this fix was intentional although not related to jdom.

dwalluck (approved):

core/src/main/java/org/commonjava/maven/ext/core/state/RelocationState.java
rnc: Yes, it was intentional.  And I agree it's not related but thought it trivial enough to include.

PR #877, opened 2021-08-26
NCLSUP-257: Document operation

dwalluck:


rnc:
As we discussed in our meeting I think this needs expanding to satisfy the original ticket.

dwalluck:
> As we discussed in our meeting I think this needs expanding to satisfy the original ticket.

Now https://github.com/release-engineering/pom-manipulation-ext/pull/885 will have more about the stages which is related to the operation in general.

dwalluck:
New version. Again, I am not sure how much of it is just a repeat of the Groovy section.

PR #878, opened 2021-08-30
Remove debug log messages

rnc:


rnc:
@dwalluck I think we both missed these :-( and they sneaked into the release. Harmless but annoying.

dwalluck (approved):

PR #883, opened 2021-09-17
NCL-6130: Add PREPARSE stage to allow running Groovy scripts without …

dwalluck:
…a project


core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -191,6 +192,10 @@ public void init( final ManipulationSession session )
     public void scanAndApply( final ManipulationSession session )
                     throws ManipulationException
     {
+        final PreparseGroovyManipulator preparseGroovyManipulator = new PreparseGroovyManipulator( null, null, pomIO );
+        preparseGroovyManipulator.init( session );
+        preparseGroovyManipulator.applyChanges( Collections.emptyList() );
+
```
dwalluck, line 198: I placed the call inside `scanAndApply` so that if called in code (and not CLI) the preparse would still be run.

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
```diff
@@ -22,15 +22,17 @@
...
 public enum InvocationStage
 {
-    FIRST(1),
-    LAST(99),
-    BOTH(Integer.MAX_VALUE);
+    PREPARSE( 0 ),
+    FIRST( 1 ),
+    LAST( 99 ),
+    BOTH( Integer.MAX_VALUE ),
+    ALL( Integer.MAX_VALUE );
```
dwalluck, line 29: So I made `ALL` with the same value as `BOTH` but I am not sure it is useful as we don't want to normally run a `BOTH` script during preparse anyway.

integration-test/src/it/setup/depMgmt5/Sample.groovy
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package org.commonjava.maven.ext.depMgmt5
+
+import org.commonjava.maven.ext.core.groovy.BaseScript
+import org.commonjava.maven.ext.core.groovy.InvocationStage
+import org.commonjava.maven.ext.core.groovy.PMEBaseScript
+import org.commonjava.maven.ext.core.groovy.InvocationPoint
+
+@InvocationPoint(invocationPoint = InvocationStage.ALL)
```
dwalluck, line 23: This whole directory is just a copy of  the `BOTH` test, so I am not sure it is necessary.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
```diff
@@ -154,20 +154,28 @@ public Translator getRESTAPI() throws ManipulationException
...
         this.modelIO = modelIO;
-        this.sessionHandler = session;
+        this.session = session;
         this.projects = projects;
+
+        if (project != null) {
+            this.gav = project.getKey();
+            this.basedir = project.getPom().getParentFile();
+        } else {
+            this.basedir = session.getPom().getParentFile();
```
rnc, line 171: `baseDir` should the same from either of these so this seems redundant (if its not the same I think we have a bug!)

Aside : formatting?

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
rnc: We should look at a way to deprecate usage of BOTH - detect and warn?

core/src/main/java/org/commonjava/maven/ext/core/impl/FinalGroovyManipulator.java
```diff
@@ -50,15 +51,15 @@ public FinalGroovyManipulator(ModelIO modelIO, FileIO fileIO, PomIO pomIO)
...
+     * for later.
      */
     @Override
     public void init( final ManipulationSession session ) throws ManipulationException
     {
-        GroovyState gs = new GroovyState( session.getUserProperties() );
+        final State state = new GroovyState( session.getUserProperties() );
         this.session = session;
-        session.setState( gs );
+        session.setState( state );
```
rnc, line 62: Any particular reasons for the renaming ... :-)

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * {@link Manipulator} implementation that can resolve a remote groovy file and execute it on executionRoot. Configuration
+ * is stored in a {@link GroovyState} instance, which is in turn stored in the {@link ManipulationSession}.
+ */
+@Named("groovy-injection-preparse")
+@Singleton
```
rnc, line 50: I don't _think_ this needs those annotations?

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+        this.session = session;
+        session.setState( state );
+
+        try
+        {
+            final GalleyInfrastructure infra = new GalleyInfrastructure( session, new DefaultMirrorSelector() );
+            infra.init( null, null, Files.createTempDirectory( "pme-" ).resolve( "cache-dir" ).toFile() );
+            final GalleyAPIWrapper wrapper = new GalleyAPIWrapper( infra );
+            modelIO = new ModelIO( wrapper );
+            fileIO = new FileIO( infra );
```
rnc, line 79: Why are you creating fileIO here? ; its passed in isn't it?

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+        {
+            throw new ManipulationException( e.getMessage(), e );
+        }
+    }
+
+    /**
+     * Apply the groovy script changes.
+     */
+    //@Override
+    public Set<Project> applyChanges( final List<Project> projects )
```
rnc, line 91: I guess this is work in progress as the projects won't be passed in (and don't exist) and the override doesn't apply

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+        {
+            final GroovyShell shell = new GroovyShell();
+
+            try
+            {
+                final Script script = shell.parse(groovyScript);
+                final InvocationPoint invocationPoint = script.getClass().getAnnotation(InvocationPoint.class);
+                final InvocationStage stage = invocationPoint != null ? invocationPoint.invocationPoint() : null;
+
+                if ( InvocationStage.PREPARSE == stage )
```
rnc, line 114: Why are lines 106->114 needed as can't `applyGroovyScript` be used? Possibly the info logger in that apply function needs to be a bit cleverer though.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
rnc: While initially I did think the CLI was a _more_ appropriate place, thinking on it further, if its in `scanAndApply` then it will work whether PME is run as a CLI **or** as an extension.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -191,6 +192,10 @@ public void init( final ManipulationSession session )
     public void scanAndApply( final ManipulationSession session )
                     throws ManipulationException
     {
+        final PreparseGroovyManipulator preparseGroovyManipulator = new PreparseGroovyManipulator( null, null, pomIO );
+        preparseGroovyManipulator.init( session );
+        preparseGroovyManipulator.applyChanges( Collections.emptyList() );
```
rnc, line 197: Given this manipulator **isn't** following the standard pattern (else as we discussed it would be accidentally run during the standard execution), what about making a simpler API ; i.e. merging `init` and `apply` methods?

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
```diff
@@ -191,6 +192,10 @@ public void init( final ManipulationSession session )
     public void scanAndApply( final ManipulationSession session )
                     throws ManipulationException
     {
+        final PreparseGroovyManipulator preparseGroovyManipulator = new PreparseGroovyManipulator( null, null, pomIO );
```
rnc, line 195: Manual creation means the injection won't work. I think this needs changing. This also affects my comments further down regarding your manual creation of fileIO etc.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * {@link Manipulator} implementation that can resolve a remote groovy file and execute it on executionRoot. Configuration
```
rnc, line 46: Minor : technically not a org.commonjava.maven.ext.core.impl.Manipulator implementation :-)

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+            final GalleyInfrastructure infra = new GalleyInfrastructure( session, new DefaultMirrorSelector() );
+            infra.init( null, null, Files.createTempDirectory( "pme-" ).resolve( "cache-dir" ).toFile() );
+            final GalleyAPIWrapper wrapper = new GalleyAPIWrapper( infra );
+            modelIO = new ModelIO( wrapper );
+            fileIO = new FileIO( infra );
+        }
+        catch ( IOException e )
+        {
+            throw new ManipulationException( e.getMessage(), e );
+        }
```
rnc, line 84: I think both of these above catch clauses could be removed if just `applyGroovyScripts` is used?

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
```diff
@@ -0,0 +1,134 @@
...
+
+                if ( InvocationStage.PREPARSE == stage )
+                {
+                    applyGroovyScript( Collections.emptyList(), null, groovyScript );
+                }
+            }
+            catch ( IOException e )
+            {
+                throw new ManipulationException( e.getMessage(), e );
+            }
```
rnc, line 122: I think both of these above catch clauses could be removed if just `applyGroovyScripts` is used?

core/src/main/java/org/commonjava/maven/ext/core/impl/FinalGroovyManipulator.java
dwalluck: No, but I figured that `setState` makes more sense to take ` `state` parameter. I can change it back if you want.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
dwalluck: Agreed.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: I am pretty sure they are `null` here since they aren't injected. The `pomIO` is passed through the parent class, though.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: If it's not going to be a `Manipulator`, then I may be able to remove all annotations related to that.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
dwalluck: I think sometimes session pom is `null` though so that can't be called unconditionally. It works since in one one case we don't have project we do have the session.

integration-test/src/it/setup/depMgmt5/Sample.groovy
rnc: I agree! I would really like to avoid adding more things to setup unless its really necessary. I would think you don't need much setup surely? Could all this be moved into the actual test?

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
rnc: I think that needs to be narrowed down? - i.e. when it might be null?

core/src/main/java/org/commonjava/maven/ext/core/impl/FinalGroovyManipulator.java
rnc: No its fine, I was just curious!

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
rnc: Yes well, see my prior comments about how the injection needs to be fixed so none of this is required.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
dwalluck: It is actually `session.getPom()`. The `TestUtil.createSession()` method does not set a pom. In most cases, there is a `File rootproject` to use for a pom, but in some cases there is none.

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
dwalluck: Now, I wonder if `ALL` even makes sense since we don't want to run `PREPARSE` (which I would think would be included in `ALL`). Really, I think 'BOTH' meaning `FIRST` and `LAST` is OK, then.

rnc: To be honest I think we have to be careful not to make assumptions on when people might want to run it. Hence wanting to deprecate BOTH and replace with ALL.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScript.java
rnc: If its test code creating not fully populated object we should fix that rather than adding if-blocks here in my opinion

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
dwalluck: I may be missing something.

If we make `BOTH` equivalent to `ALL` and don't change the semantics, then the name is actually misleading.

If we change the semantics, then scripts which currently work with `BOTH` are likely to break with `PREPARSE` where (`ALL = { PREPASE, FIRST, LAST }`) since it's missing, e.g., the project information.

Yet, we currently have no other way of supporting any other combinations of stages I thought.

integration-test/src/it/setup/depMgmt5/Sample.groovy
dwalluck: I will try. I think the reason was I was looking for an existing script with `BOTH` and the only one I found was in setup.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: It was currently to block execution on `stage != PREPARSE` since the `if` inside `applyGroovyScript` currently runs with `BOTH`, but as I explained elsewhere, running `PREPARSE` in that case is going to break existing tests.

core/src/main/java/org/commonjava/maven/ext/core/ManipulationManager.java
dwalluck: Sure.

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
rnc: I know the `BOTH` name is misleading if we make it equivalent to `ALL`. Equally though its misleading now there are  _three_ states and not two. A user can still check what stage the script is in before choosing to continue processing or not. 
I do not believe `BOTH` is appropriate any more ; a more radical solution is simply to remove it but if you were to choose to do that I would suggest checking with users first?

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
```diff
@@ -90,24 +91,38 @@ public static Model getDummyModel ()
         return result;
     }
 
+    // Public API for manipulator-examples-project
+    public static ManipulationSession createSession( Properties p, File pom ) throws ManipulationException
+    {
+        return createSessionAndManager( p, pom ).getSession();
+    }
```
rnc, line 98: Why is this been created?

dwalluck: Per our discussion, since you want to use `session.getPom()` instead of using the project, then the pom `File` must be passed into the session. Currently, sessions are being created with a `null` pom in `createSession`. You had said to fix the tests rather than have that `if` statement so that's why I added a method to take a `File`.

rnc: Ah right, the `Public API for manipulator-examples-project` confused me. Can you do me a favour and layer the commits as its impossible to see the changes between subsequent pushes and I have to re-review everything again - which is rather time consuming. You can also squash at the end once its complete.

integration-test/src/it/setup/depMgmt5/Sample.groovy
rnc: That shouldn't preclude moving every to a test ; otherwise we are imposing the cost of a setup for no apparent reason.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: Can we alter the comment please

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
```diff
@@ -129,18 +145,21 @@ public static SMContainer createSessionAndManager( Properties p ) throws Manipul
         }
         final MavenSession mavenSession = new MavenSession( container, null, req, new DefaultMavenExecutionResult() );
 
+        mavenSession.getRequest().setPom( pom );
```
rnc, line 148: What happens if pom is null?

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
```diff
@@ -175,9 +194,9 @@ public static Object executeMethod( Object instance, String name, Class<?>[] typ
         @Getter
         private final MavenExecutionRequest request;
 
-        private SMContainer( MavenExecutionRequest req, ManipulationSession session, ManipulationManager manager )
+        private SMContainer( MavenExecutionRequest request, ManipulationSession session, ManipulationManager manager )
```
rnc, line 197: Now that the pom _can be_ passed to `createSessionAndManager` I don't think the request needs to be accessible via the `SMContainer`. While this is used in PME tests and in the groovy-example project I think this API change would make things simpler.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
```diff
@@ -90,24 +91,38 @@ public static Model getDummyModel ()
...
     @SuppressWarnings( "deprecation" )
-    public static SMContainer createSessionAndManager( Properties p ) throws ManipulationException
+    public static SMContainer createSessionAndManager( Properties  p ) throws ManipulationException
+    {
+        return createSessionAndManager( p, null );
+    }
+
+    // Public API for manipulator-examples-project
+    @SuppressWarnings( "deprecation" )
+    public static SMContainer createSessionAndManager( Properties p, File pom ) throws ManipulationException
```
rnc, line 115: Can we add Javadoc for this and the preceding method

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
rnc: Can you point me to an example ? Can't the tests only run with first/last phases?

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
dwalluck: I am not sure yet if this change breaks manipulator-examples-project and then we'd have to switch to the new API.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
dwalluck: Setting a `null` pom should be the same as the default (unset) and seems to be OK as long as you don't intend you use it.

The code you wanted `session,getPom().getParentFile()` requires a non-null pom. In that case, for the few tests that didn't use a pom. I just set it to a temp file which makes this call work to get the parent directory. Even though the pom itself is not valid. it's never called.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
dwalluck: After this API s finalized, I will submit a pull request to change the examples (if necessary).

integration-test/src/it/setup/depMgmt5/Sample.groovy
dwalluck: I moved the test out of setup.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: On the one hand, we manually instantiate the class, but on the other, we are still using injection here.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: I've removed the `if` from this class, so now it should be a question of whether the `if` in `BaseGroovyManipulator` looks correct.

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
dwalluck: I have tried to keep `BOTH` in the `if` in `BaseGroovyManipulator` (by checking `FIRST` and `LAST`) if it looks right to you. In which case, I am not sure it really needs to be deprecated, we'd basically be keeping the original meaning while adding `ALL` to cover the third state (or beyond).

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: Done.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
dwalluck: There's no javadoc on most `public` methods here. Should we add it?

I still don't know if this is the right API. In some cases, we create a session without a POM. For example, a `ModelIO` requires a session and is generally used for resolving files outside of a project. The old version of the code seemed to handle the case of not having a file.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
rnc: I didn't think you were manually instantiating it any more?

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: I think we're talking at cross-purposes - even within _here_ I don't think the request needs to be accessible due to your preceding changes?

dwalluck: I made the change to `SMContainer` since `request` is no longer needed to set the pom as you said. Still, I think we need a better solution for when there is no pom.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
dwalluck: I am manually instantiating the class itself initially before it gets passed to `ManipulationManager`. But, there is an `@Inject` on the constructor so that the  constructor parameters get set automatically, so I think using `@Named` on the class itself is a good idea now?

rnc: Can you point me where? I thought you were relying on injection. (Which means the above should stay)

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: You mean whether manip-examples needs to use this API?

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
rnc: I'm honestly not sure of the benefit to keeping `BOTH` as a differentiator from `ALL` ; plus it will make your upcoming documentation modifications more complex.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: I suppose we _should_ add it ; we haven't been especially diligent

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: Ok.

core/src/main/java/org/commonjava/maven/ext/core/impl/PreparseGroovyManipulator.java
rnc: Discussed in meeting ; this class is using injection so we do need these.

core/src/test/java/org/commonjava/maven/ext/core/fixture/TestUtils.java
rnc: Agreed to leave it for now as its test code.

core/src/main/java/org/commonjava/maven/ext/core/groovy/InvocationStage.java
dwalluck: We have decided to remove `BOTH`. However, I will need to check as `BOTH` is used in one of the IT setup tests shared by some other tests.

dwalluck: Removed `BOTH`, changed existing IT to use `ALL` instead.

rnc (approved):

PR #884, opened 2021-10-01
Add support for building under Java 17

dwalluck:


integration-test/src/it/rest-dependency-version-manip-child-module/test.properties
```diff
@@ -18,5 +18,6 @@ restURL=http://127.0.0.1:8089
 versionIncrementalSuffix=redhat
 overrideTransitive=false
 strictAlignment=true
+dependencyOverride.org.jboss\:jboss-parent@*=37.0.0.redhat-00001
```
dwalluck, line 21: I had to force the version or else it gets `37-redhat-1`. Changing `versionIncrementalSuffixPadding` didn't seem to help. Wasn't sure how to set the properties in order to get `37 --> 37.0.0.redhat-00001`

dwalluck: Obviously, the `dependencyOverride` works, but I thought it would be possible without this by changing something else.

dwalluck: I assume we want the upstream version, which is `37`, not `37.0.0`. Maybe I was wrong thinking the transformation was `37 --> 37.0.0.redhat-00001` and not `37.0.0 --> 37.0.0.redhat-00001`. But, again, it isn't using the suffix padding either.

dwalluck: Indeed, I believe the reason for the failure is that no such version `37-redhat-1` exists. I was just wondering if there was a better way to cause it to generate `37.0.0.redhat-00001` from the value `37`. As you said, the previous version had both `11` (community) and `11-redhat-1` existing in the repos.

The reason I made a big deal out of this is I thought I may be missing something about how it works. For example, `versionIncrementalSuffixPadding` had no effect and it still put a single `1`. Is it hardcoding the response somewhere?

PR #885, opened 2021-10-01
NCL-6130: Document invocation stages

dwalluck:


guide/groovy.md
```diff
@@ -103,15 +94,26 @@ import org.commonjava.maven.ext.core.groovy.InvocationPoint
...
+|------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `PREPARSE` | Runs the script _prior to parsing any POM files_ and before any manipulations have been performed, thus allowing the modification of POM files on disk before they are read into memory. |
+| `FIRST`    | Runs the script _first_ after all POM files have been read into memory, but before any manipulations have been performed.                                                                |
+| `LAST`     | Runs the script _last_ after all modifications to the in-memory POM files have been performed.                                                                                           |
+| `ALL`      | Runs the during _all_ possible stages: `PREPARSE`, `FIRST`, and `LAST`.  The `getInvocationStage()` API can be used to determine in which stage the script is currently running.         |
```
rnc, line 115: I think we need to make it clearer the consequences of editing files on disk during FIRST/LAST i.e. the warning from above. Otherwise I like the new section

dwalluck: Do you just want to move the warning section down here? I probably just deleted it and forgot to move it.

dwalluck: I think it may make sense as a paragraph now instead of a straight warning since it is safe now in the `PREPARSE` phase, which is somewhat explained in the table, but could be explained more later.

rnc: I think it needs more than just moving the section ; I agree you might need to reconsider how to structure it given the warning applies to some sections but not every one.

guide/groovy.md
```diff
@@ -103,15 +93,31 @@ import org.commonjava.maven.ext.core.groovy.InvocationPoint
...
+
+It is safe to modify POM files on disk during the `PREPARSE` stage. However, if you modify a POM file on disk during any
+other stage, the modifications will be overwritten when the in-memory POM file is written back out to disk. To alter a
+POM file in memory, call `Project.getModel()` to  retrieve the `org.apache.maven.model.Model` instance and modify
+that instead, .e.g., `pme.getProject().getModel().setVersion( "1.0.0" )`.
```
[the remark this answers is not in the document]
dwalluck: Done.

rnc (approved):

PR #886, opened 2021-10-01
NCL-6614: Use table layout for API

dwalluck:


dwalluck:
The previous version of the docs said that `void inlineProperty(Project, String)` was removed and replaced by `void inlineProperty(Project, ProjectRef)` in 3.8. However, if I am not mistaken, both seem to exist.

rnc:
> The previous version of the docs said that `void inlineProperty(Project, String)` was removed and replaced by `void inlineProperty(Project, ProjectRef)` in 3.8. However, if I am not mistaken, both seem to exist.

No, the original had a String representing `groupId:artifactId` while the new String form represents a _name_ of a property. See the code in BaseScriptUtils for more details. I actually think the original description that had `Supports * as a wildcard for artifactId.` is incorrect - that doesn't appear in the javadoc either.

guide/groovy.md
```diff
@@ -117,79 +117,27 @@ The API can then be invoked by e.g.
...
+| void inlineProperty([Project](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/model/Project.java), [String](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html)) | Allows the specified property to be inlined in any dependencies/dependencyManagement. This is useful to split up properties that cover multiple separate projects. Supports `*` as a wildcard for artifactId. | 3.5 |
+| void inlineProperty([Project](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/model/Project.java), [ProjectRef](https://github.com/Commonjava/atlas/blob/master/identities/src/main/java/org/commonjava/atlas/maven/ident/ref/ProjectRef.java)) | Allows the specified group:artifact property to be inlined in any dependencies/dependencyManagement. This is useful to split up properties that cover multiple separate projects. Supports `*` as a wildcard for artifactId. | 3.8 |
```
rnc, line 132: See my main comment replying to yours where I think this latter sentence might be incorrect

guide/groovy.md
```diff
@@ -117,79 +117,27 @@ The API can then be invoked by e.g.
...
-        <tr>
-        <td><code>lookupProjectVersions</code></td>
-        <td>Lookup versions (e.g. for a project) ignoring DA suffix priority schemes returning the latest version</td></tr>
-    </table>
-</td>
-</tr>
-</table>
+| Method | Description | Since      |
+| -------|:------------|:-----------|
+| [Map](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html)<[ProjectVersionRef](https://github.com/Commonjava/atlas/blob/master/identities/src/main/java/org/commonjava/atlas/maven/ident/ref/ProjectVersionRef.java), [String](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html)> translateVersions([List](https://docs.oracle.com/javase/7/docs/api/java/util/List.html)<[Project](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/model/Project.java)>) | Translate the versions. _Removed in 4.3; use `lookupProjectVersions` or `lookupVersions` instead_. | 1.7 |
```
rnc, line 122: This is an interesting one - how to deal with _removed_ API. I am not sure whether its better to have it in or remove it. Or use 'strike-out' to emphasise its no longer available?

guide/groovy.md
```diff
@@ -117,79 +117,27 @@ The API can then be invoked by e.g.
...
+| [Map](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html)<[ProjectVersionRef](https://github.com/Commonjava/atlas/blob/master/identities/src/main/java/org/commonjava/atlas/maven/ident/ref/ProjectVersionRef.java), [String](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html)> lookupVersions([List](https://docs.oracle.com/javase/7/docs/api/java/util/List.html)<[Project](https://github.com/release-engineering/pom-manipulation-ext/blob/master/common/src/main/java/org/commonjava/maven/ext/common/model/Project.java)>) throws [RestException](https://github.com/release-engineering/pom-manipulation-ext/blob/main/io/src/main/java/org/commonjava/maven/ext/io/rest/RestException.java) | Lookup versions respecting DA suffix priority schemes which will return the best matched version. | 4.4 |
```
rnc, line 141: I _think_ if I understand correctly your order for the table as based upon the 'since' column? Not the alphabetical name of the function?

guide/groovy.md
dwalluck: OK, I think it wasn't clear to me what that meant, since other APIs didn't have an argument name. I am wondering if it should be listed twice.

guide/groovy.md
dwalluck: Strike out the name maybe? I also thought to write a range for since like `[1.7, 4.3]`.

guide/groovy.md
dwalluck: Yes, it's in order of since version and then alphabetical. This is closer to how it was originally which was chronological.

dwalluck:
> I actually think the original description that had `Supports * as a wildcard for artifactId.` is incorrect - that doesn't appear in the javadoc either.

I think it actually supports `*` for _either_ `artifactId` _or_ `groupId` automatically. So, even for the other `inlineProperty`, it doesn't seem entirely correct.

Internally, they both just call `SimpleProjectRef.parse()` to set the values.

The actual code to check for `'*'` appears in `DependencyManipulator`.

guide/groovy.md
rnc: What should be listed twice?

guide/groovy.md
rnc: Strike-out seems reasonable and you could add the range for the 'since' as well ?

guide/groovy.md
rnc: Ok

rnc:
> > I actually think the original description that had `Supports * as a wildcard for artifactId.` is incorrect - that doesn't appear in the javadoc either.
> 
> I think it actually supports `*` for _either_ `artifactId` _or_ `groupId` automatically. So, even for the other `inlineProperty`, it doesn't seem entirely correct.
> 
> Internally, they both just call `SimpleProjectRef.parse()` to set the values.
> 
> The actual code to check for `'*'` appears in `DependencyManipulator`.


I'm really not clear what you mean - maybe ping me offline to chat about this?

guide/groovy.md
dwalluck: I meant whether that API that changed should be two rows or one. I guess one row with a note in the description?

guide/groovy.md
dwalluck: Sure. I can do both.

rnc:
I think the invocation stage might also need a since column?

guide/groovy.md
rnc: Oh you mean the old api and whether we should list it.  Hmm not sure, maybe your idea of a note is best

dwalluck:
> I think the invocation stage might also need a since column?

Sorry, I had to force push due to rebase in order to get the stages into the doc. Original states were in 3.5, and the new ones will be in 4.6. I could add a `@since` to the javadoc for the states or the APIs in a different pull request maybe.

rnc (approved):

PR #891, opened 2021-10-15
NCL-6663: Clarify use of `versionOverride` property when used without…

dwalluck:
… `versionSuffix` property

### DIFF
--- a/guide/project-version-manip.md
+++ b/guide/project-version-manip.md
@@ -51,15 +51,27 @@ If the current version of the project is "1.2.0.GA", the new version set during 
 
 ### Version override
 
-The version can be forcibly overridden by using the property `versionOverride` (*Deprecated property `version.override` for versions **3.8.1 and prior***)
+The version can be forcibly overridden by using the property `versionOverride` (*Deprecated property `version.override`
+for versions **3.8.1 and prior***)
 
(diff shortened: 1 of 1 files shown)


guide/project-version-manip.md
```diff
@@ -61,6 +61,15 @@ If the current version of the project is "6.2.0", the new version set during the
...
 
+Note that the entire value of `versionOverride` is taken to be the version, including any suffix, and may therefore be
+subject to automatic version increment. For example, running
+
+    mvn install -DversionOverride=6.1.0.Final-rebuild-1
+
+would result in the final version being "6.1.0.Final-rebuild-2" and not "6.1.0.Final-rebuild-1". Therefore, if you
+intend to override both the version and the suffix, then be sure to specify properties for each separately as shown
+above.
+
```
rnc, line 72: This is slightly confusing to read IMHO.  versionOverride is meant to override the version NOT the suffix. If you set an override with part of a suffix it gets really confusing - try it out! Really the versionOverride should _just_ be used for the version and suffix handling left to other options i.e. you need to combine options to get your ideal desired version.

rnc: Can we rephrase this please

rnc (changes requested):

PR #892, opened 2021-10-20
NCL-5477: Handle properties in `RangeResolver`

dwalluck:


core/src/main/java/org/commonjava/maven/ext/core/impl/RangeResolver.java
```diff
@@ -79,181 +83,234 @@ public void init( final ManipulationSession session )
...
+
+                            for ( final Plugin plugin : plugins )
+                            {
+                                handleVersionWithRange( plugin, pi );
+                            }
+                        }
+
+                        final List<Plugin> profilePlugins = profileBuild.getPlugins();
+
+                        for ( final Plugin plugin : profilePlugins )
```
rnc, line 191: Can you explain why all of the above has been changed?

core/src/main/java/org/commonjava/maven/ext/core/impl/RangeResolver.java
```diff
@@ -79,181 +83,234 @@ public void init( final ManipulationSession session )
...
         try
         {
             mavenMetadataView = readerWrapper.readMetadataView( ga );
         }
-        catch ( GalleyMavenException e )
+        catch ( final GalleyMavenException e )
         {
-            throw new ManipulationUncheckedException( new ManipulationException( "Caught Galley exception processing artifact", e ) );
+            final Throwable t = new ManipulationException( "Caught Galley exception processing artifact", e );
+            throw new ManipulationUncheckedException( t );
```
rnc, line 306: Why this change please?

core/src/main/java/org/commonjava/maven/ext/core/impl/RangeResolver.java
```diff
@@ -79,181 +83,234 @@ public void init( final ManipulationSession session )
...
+        catch ( final InvalidVersionSpecificationException | ManipulationException e )
         {
             throw new ManipulationUncheckedException( new ManipulationException( "Invalid range", e ) );
         }
     }
 
-    private void handleVersionWithRange( Dependency d )
+    private void handleVersionWithRange( Dependency dependency, PropertyInterpolator pi ) throws ManipulationException
     {
+        final String version = pi.interp( dependency.getVersion() );
```
rnc, line 256: `PropertyResolver` already does this and handles properties defined within a parent (but still within the project)

There is also the functions on Project that can return a fully resolved list of Deps/Plugins/etc handling resolving as well.

I think we should reuse some of that instead which would also mean you wouldn't need to pass a `PropertyInterpolator` around and simplify some of these changes.

PR #895, opened 2021-11-19
NCLSUP-473: Fix issues with `overrideTransitive` option

dwalluck:


integration-test/src/it/remote-dependency-management-project-version/verify.groovy
```diff
@@ -14,13 +14,12 @@
...
 def pom = new XmlSlurper().parse( pomFile )
 
-def dependency = pom.dependencyManagement.dependencies.dependency.find { it.artifactId.text() == "commons-lang" }
-assert dependency != null
-assert dependency.version.text().endsWith ("{project.version}")
+def dependency = pom.dependencyManagement.dependencies.dependency.find { it.artifactId.text() == 'commons-lang' }
+assert dependency.version.text() == '\${project.version}'
 
 def buildLog = new File( basedir, 'build.log' )
-assert buildLog.text.contains("so skipping")
```
dwalluck, line 26: OK, so this test is "lost" since this artifact is already removed by the new code.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -390,12 +391,13 @@ private void apply( final Project project, final Model model, final Map<Artifact
...
-            if ( commonState.isOverrideTransitive() )
+            if ( commonState.isOverrideTransitive() && dependencyState.getRemoteBOMDepMgmt() != null )
             {
-                final List<Dependency> extraDeps = new ArrayList<>();
+                final Collection<ArtifactRef> overrideRefs = overrides.keySet();
+                final Collection<Dependency> extraDeps = new ArrayList<>( overrideRefs.size() );
 
                 // Add dependencies to Dependency Management which did not match any existing dependency
-                for ( final ArtifactRef pvr : overrides.keySet() )
+                for ( final ArtifactRef pvr : overrideRefs )
```
rnc, line 400: Why can't overridesRefs be inlined to `overrides.keySet()` ?

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -430,6 +432,11 @@ private void apply( final Project project, final Model model, final Map<Artifact
                     dependencyManagement.getDependencies().addAll( 0, extraDeps );
                 }
             }
+            else if ( commonState.isOverrideTransitive() && dependencyState.getRemoteBOMDepMgmt() == null )
+            {
+                logger.warn( "Ignoring {}={} since it was used without the {} option", "overrideTransitive",
+                        commonState.isOverrideTransitive(), "dependencyManagement" );
```
rnc, line 438: Why are we parametrising a hard-coded string?  What about:
```
 logger.warn( "Ignoring overrideTransitive=true since it was used without the dependencyManagement option");
```

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -626,19 +633,25 @@ else if ( strict && oldVersion.contains( "$" ) &&
...
         {
-            Entry<ArtifactRef, String> e = it.next();
+            final Entry<ArtifactRef, String> e = it.next();
+
             session.getProjects().forEach( p -> {
-                if ( e.getKey().asProjectVersionRef().equals( new SimpleProjectVersionRef( p.getGroupId(), p.getArtifactId(), p.getVersion() ) ) )
+                final ProjectRef currentOverride = e.getKey().asProjectRef();
+                final ProjectRef projectGA = new SimpleProjectRef( p.getGroupId(), p.getArtifactId() );
+
+                if ( currentOverride.equals( projectGA ) )
```
rnc, line 652: Thinking on this we could probably make it simpler by just doing:
```
if ( e.getKey().getGroupId().equals(p.getGroupId()) && e.getKey().getArtifactId().equals(p.getArtifactId() )
{
   logger.warn....
```

This avoids creating the new objects and is a bit simpler.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: I was hoping that there were constants somewhere that gave the properties for the options and then I would fill them in from that. But I guess not? Or maybe they're not `public`?

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: I pulled `overrideRefs = overrides.keySet()` out of the loop in order to make a small optimization from

```
final List<Dependency> extraDeps = new ArrayList<>();
```

to

```
final Collection<Dependency> extraDeps = new ArrayList<>( overrideRefs.size() );
```

since we know it's (maximum) size ahead of time can only be equal to the size of `overrides`.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: Done in 90013dc.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: Please look at dc6017d to see what I would like to do.

The `true` part could be hardcoded or even left out, but I think that these property names should all be `public static final` constants so that they can be referenced elsewhere in the code. I checked and some are `public`, but most are `private` to that class.

It's probably best done as a separate commit. I changed only the two classes I needed for this commit.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
rnc: Default capacity is 10 I believe and it increases in O(n) time. So whether in reality this causes an actual improvement is doubtful. However, its harmless enough so we could keep it.

core/src/main/java/org/commonjava/maven/ext/core/state/CommonState.java
```diff
@@ -38,28 +38,28 @@
...
-    private static final String TRANSITIVE_OVERRIDE_PROPERTY = "overrideTransitive";
+    public static final String TRANSITIVE_OVERRIDE_PROPERTY = "overrideTransitive";
 
     /**
      * When true, clashes with cached properties will throw an exception in PropertyResolver. Setting this to false will prevent
      * that. Default value is true.
      */
     @ConfigValue(docIndex = "dep-manip.html#property-clash-replacement")
-    private static final String PROPERTY_CLASH_FAILS = "propertyClashFails";
+    public static final String PROPERTY_CLASH_FAILS = "propertyClashFails";
```
rnc, line 48: Why are we changing these from private to public?

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
```diff
@@ -430,6 +432,12 @@ private void apply( final Project project, final Model model, final Map<Artifact
                     dependencyManagement.getDependencies().addAll( 0, extraDeps );
                 }
             }
+            else if ( commonState.isOverrideTransitive() && dependencyState.getRemoteBOMDepMgmt() == null )
+            {
+                logger.warn( "Ignoring {}={} since it was used without the {} option",
+                        CommonState.TRANSITIVE_OVERRIDE_PROPERTY, commonState.isOverrideTransitive(),
+                        DependencyState.DEPENDENCY_MANAGEMENT_POM_PROPERTY );
```
rnc, line 439: I still think that parametrising a string to combine two strings in the end is rather pointless and seems unnecessary - and potentially it less performant than a single string. I don't think this needs any parameters.

core/src/main/java/org/commonjava/maven/ext/core/state/DependencyState.java
```diff
@@ -52,13 +51,13 @@
     @SuppressWarnings("DeprecatedIsStillUsed")
     @Deprecated
     @ConfigValue( docIndex = "dep-manip.html#exclusions-and-overrides", deprecated = true )
-    private static final String DEPENDENCY_EXCLUSION_PREFIX = "dependencyExclusion.";
+    public static final String DEPENDENCY_EXCLUSION_PREFIX = "dependencyExclusion.";
```
rnc, line 54: As per above, why change from private to public?

dwalluck: This was in response to your comment about a "hard-coded string" which I replied to. The strings were not meant to be hardcoded in the sense that I wanted to use constants for those.

I think it should be possible (and preferred) in the code to refer to the names of properties by constants instead of having to manually lookup or remember what is being used and hardcode that string into the messages. They are already there, but not all are made public.

Property constants which are currently public are mostly used in tests only which means they're not being used in messages even though they are being correctly used for property lookup.

Public but unused:

`BOMInjectingState`: all
`RelocationState.PLUGIN_RELOCATIONS`
`RESTState`: all except  `RESTState.REST_URL`  which is used for value lookup in `Cli` log message

Public and used in tests:

`CommonState.STRICT_ALIGNMENT_IGNORE_SUFFIX`
`CommonState.EXCLUDED_SCOPES`
`DistributionEnforcingState`: all
`RelocationState.DEPENDENCY_RELOCATIONS`
`VersioningState`: all except `VersioningState.VERSION_OVERRIDE_SYSPROP` seems unused

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: True, this list in particular is probably so small as to not matter. I imagine the majority of overrides will fall under 10 anyway, or at least between 10 and 20.

I guess it's more of a question of what the general policy should be. My feeling is that if we know a list's size ahead of time, we should definitely put the size. If we just know its upper bound, I guess that's more debatable. This bound means that it won't have to grow, which I understand is the more expensive operation vs. using a little bit of extra memory at the outset.

integration-test/src/it/remote-dependency-management-project-version/verify.groovy
dwalluck: This actual case is tested later by the new test, because the dependency is no longer present by the time it hits this code. The first case of `if ( oldVersion.equals( Version.PROJECT_VERSION ) )` should no longer be possible I think.

The second case `if ( oldVersion.contains( "$" ) && project.getVersion().equals( resolvedValue ) ) )` I am not sure.

https://github.com/release-engineering/pom-manipulation-ext/blob/87293b28709a3ddb20df71f7ff8a846159f2cb6b/core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java#L524

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
rnc: Agreed, as a general policy I guess if we have the information we can use it.

core/src/main/java/org/commonjava/maven/ext/core/state/DependencyState.java
rnc: Ok, I'm not particularly bothered about making them public given that they static consants, I more curious on the reason for the change

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
rnc: Pros/cons - you probably negate any performance change in your work elsewhere via the extra string 'appends'. However, there is an argument to use the reference to the constant thereby eliminating any issues with future renaming. I guess any impact is minimal though I would still like to see the boolean parameter got rid of.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
rnc: [ see conversation above ]

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
dwalluck: Should it even have the "=true" in the message or just "Ignoring {} option"? I used the word option but I wasn't sure if that's what we're calling them (or "property" might be better).

core/src/main/java/org/commonjava/maven/ext/core/state/DependencyState.java
dwalluck: The reason is to be able to access the properties (map keys) as constants.

I went through the code to find why some are public and some aren't. It appears that if a test was written, then it was made public.

I would like to submit a separate PR to make them all public and find any hardcoded uses after this one.

core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java
rnc: Yeah you could remove the =true I guess.
How about 'flag' ? I think option is actually a bit better than property ; its a setting on PME after all.

dwalluck: Done in f632569.

core/src/main/java/org/commonjava/maven/ext/core/state/DependencyState.java
rnc: Yes, pretty much - I've kept things as self-contained and private until they needed to be revealed.

integration-test/src/it/remote-dependency-management-project-version/verify.groovy
rnc: I'm not entirely sure. I think that part in DepManipulator was changed in PRODTASKS-464. Given that we now remove group:artifact from the potential list, ignoring whether the version might be a property or not I think this block won't be hit. I think it would be good to verify this so we can remove this block if we are sure.

dwalluck: To make a long story short, I think the current situation is fine and we just need to remove this block of code.

It seems that the test was added at the same time as this condition in https://github.com/release-engineering/pom-manipulation-ext/pull/570. Since the project G:A is removed from the list, this condition can no longer be hit, regardless of what the version is there (project version or property). It's effectively already tested by that very same test from that PR, but instead it prints the new message from this PR.

This being the case, I think we should update this PR to remove the entire block https://github.com/release-engineering/pom-manipulation-ext/blob/87293b28709a3ddb20df71f7ff8a846159f2cb6b/core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyManipulator.java#L524-L528.

We can leave the old test there with the one-line change for the log message as it is now although there is probably some overlap now between the old test and the new ones.

rnc: Ok thats fine by me if you're sure that block isn't required now.

dwalluck: I am going to modify the test to use a GA != the project GA and see what happens.

dwalluck: In 63167c1, I switched the dependency from commons-lang to commons-net so it doesn't match project the project GA. This means the block is hit now in the test again. I noticed the block there never checks GA match, only V.

rnc (approved):

PR #896, opened 2021-11-24
Resolve issues with using as a pom extension

### LINKED ISSUE #832: Project version manipulation not working (?!?)

rnc:
Fixes #832 

See https://issues.apache.org/jira/browse/MNG-7160

Thanks to @gnodet

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
 
 POM Manipulation Extension (PME) is a Maven tool to align the versions in your POMs according to some external reference, sort of like a BOM but much more extensive and without the added baggage of a BOM declaration.
 
-It is supplied as a core library, a Maven extension (in the sense of installing to `lib/ext`, not [pom extensions](https://maven.apache.org/pom.html#Extensions)) and a command line tool.
+It is supplied as a core library, a Maven extension (either by installing to `lib/ext`, or using as a [pom extensions](https://maven.apache.org/pom.html#Extensions)) and a command line tool.
 
 This extension combines many of the features of [VMan](https://github.com/jdcasey/pom-version-manipulator), [Maven Versioning Extension](https://github.com/jdcasey/maven-versioning-extension) and [Maven Dependency Management Extension](https://github.com/jboss/maven-dependency-management-extension).
 
--- a/ext/pom.xml
+++ b/ext/pom.xml
@@ -33,6 +33,12 @@
     <dependency>
       <groupId>org.commonjava.maven.ext</groupId>
       <artifactId>pom-manipulation-common</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>org.codehaus.plexus</groupId>
+          <artifactId>plexus-utils</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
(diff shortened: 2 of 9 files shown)


dwalluck:
It looks good to me. It looks like we "just" had to exclude all instances of plexus-utils from our classpath.

I just wanted to be clear on two things said on https://issues.apache.org/jira/browse/MNG-7160:

1. That "this will only work when the extension does not actually need plexus-utils classes other than the XPP3 ones exported from the maven api realm". I guess we just get lucky there, otherwise it would not be possible?
2. That "this state means that it's always possible to use the same extension as a core extension or as a build extension, which I don't think is a good state". So is this a bug or a feature? In other words, even though we have a workaround, I suppose we could have left it as only "officially" supported as a build extension.

gnodet:
> It looks good to me. It looks like we "just" had to exclude all instances of plexus-utils from our classpath.
> 
> I just wanted to be clear on two things said on https://issues.apache.org/jira/browse/MNG-7160:
> 
> 1. That "this will only work when the extension does not actually need plexus-utils classes other than the XPP3 ones exported from the maven api realm". I guess we just get lucky there, otherwise it would not be possible?
> 2. That "this state means that it's always possible to use the same extension as a core extension or as a build extension, which I don't think is a good state". So is this a bug or a feature? In other words, even though we have a workaround, I suppose we could have left it as only "officially" supported as a build extension.

Sorry, I meant "it's not always possible", I fixed the comment...  The goal is to reconcile core and build extensions when possible.  The https://github.com/apache/maven/pull/616 PR offers some ways to configure the classloading strategy for core extensions in way which is similar to build extensions, so this bridges the classloading gap.

dwalluck:
> Sorry, I meant "it's not always possible", I fixed the comment... The goal is to reconcile core and build extensions when possible. The [apache/maven#616](https://github.com/apache/maven/pull/616) PR offers some ways to configure the classloading strategy for core extensions in way which is similar to build extensions, so this bridges the classloading gap.

OK, thanks. We had only been using the extension via the POM method before this PR, so I just wanted to make sure we weren't doing anything out of the ordinary.

dwalluck (approved):

PR #909, opened 2022-01-25
NCLSUP-531 Improve documentation on endpoints

rnc:


guide/dep-manip.md
```diff
@@ -91,36 +82,6 @@ The lookup REST endpoint should follow:
 </tr>
```
dwalluck, line 82: I Think we might be missing a `</td>` tag here.

dwalluck (changes requested):

guide/dep-manip.md
rnc: I count 2 open td and 2 close? Where please?

dwalluck: Nevermind, it seems that 2 lines were folded on the left side 89--90 and it was there.

dwalluck (approved):

PR #944, opened 2022-10-18
NCLSUP-760 Add ability to add injected dependencies to maven-dependen…

rnc:
…cy-plugin


ppatierno (approved):
Thanks for this addition!

PR #947, opened 2022-10-19
NCLSUP-779 Resolve embedded property update issues

rnc:


rnc:
@djcoleman Feel free to review to check this covers your use-cases please.

djcoleman (approved):

djcoleman:
I'm not exactly sure what the change does, but I tried it on camel-quarkus and the version now looks good. That's the only use case I know about. Thanks for fixing it.

PR #951, opened 2022-11-19
[NCLSUP-799] extend inline property feature to plugins section

rgdoliveira:
Jira: https://issues.redhat.com/browse/NCLSUP-799

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScriptUtils.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScriptUtils.java
@@ -73,6 +73,15 @@ public abstract class BaseScriptUtils extends Script implements MavenBaseScript
                               a.getValue().setVersion(
                                               PropertyResolver.resolvePropertiesUnchecked( getSession(), currentProject.getInheritedList(), a.getValue().getVersion() ) );
                           } );
+            currentProject.getResolvedPlugins( getSession() )
+                          .entrySet().stream()
+                          .filter( a -> ( groupArtifact.getArtifactId().equals( "*" ) && a.getKey().getGroupId().equals( groupArtifact.getGroupId()) ) ||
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -635,6 +636,7 @@
       <plugin>
         <groupId>org.projectlombok</groupId>
         <artifactId>lombok-maven-plugin</artifactId>
+        <version>${lombokPluginVersion}</version>
```
rgdoliveira, line 639: Unit Test `testInlineProperty` uses this pom.xml file to run the tests, that's why I created the `lombokPluginVersion` property.

core/src/main/java/org/commonjava/maven/ext/core/groovy/BaseScriptUtils.java
```diff
@@ -73,6 +73,15 @@ public void inlineProperty ( Project currentProject, ProjectRef groupArtifact )
                               a.getValue().setVersion(
                                               PropertyResolver.resolvePropertiesUnchecked( getSession(), currentProject.getInheritedList(), a.getValue().getVersion() ) );
                           } );
+            currentProject.getResolvedPlugins( getSession() )
```
rnc, line 76: I wonder if we should cover managed plugins as well as plugins (given that the function currently covers dependencies and managed dependencies) ?

rgdoliveira: Good point, I will try to add managed plugins as well.

rgdoliveira: @rnc I implemented this change and pushed a new commit to the PR.

rnc (approved):

rgdoliveira:
It seems both CI for maven 3.3.9 and 3.5.4 failed due to time out when downloading artifacts.

rnc:
Thanks for the PR!

PR #966, opened 2023-01-25
update java version

huzefab:


rnc:
Can you explain why you are changing this property?

rnc:
I'm closing this - this change will break the java versioning for various plugins.

PR #974, opened 2023-04-27
Add new fields for PNC Specific ranking/filtering of dependencies

michalovjan:
Requires for compilation: https://github.com/project-ncl/dependency-analysis/pull/561

Tested locally.

### DIFF
--- a/core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
+++ b/core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
@@ -23,7 +23,7 @@ import org.commonjava.maven.ext.core.impl.DependencyManipulator;
 import org.commonjava.maven.ext.core.util.PropertiesUtils;
 import org.commonjava.maven.ext.io.rest.DefaultTranslator;
 import org.commonjava.maven.ext.io.rest.Translator;
-import org.jboss.da.model.rest.Constraints;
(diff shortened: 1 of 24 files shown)


rnc (approved):
Looks good. When DA snapshot is available we can run the tests.

core/src/main/java/org/commonjava/maven/ext/core/state/RESTState.java
```diff
@@ -64,6 +70,27 @@ public class RESTState implements State
...
+    public static final String REST_GLOBAL_DEPENDENCY_ALLOW_LIST = "restDependencyAllowList";
+
+    @ConfigValue( docIndex = "dep-manip.html#rest-pnc-specific-properties")
+    public static final String REST_GLOBAL_DEPENDENCY_DENY_LIST = "restDependencyDenyList";
+
+    @ConfigValue( docIndex = "dep-manip.html#rest-pnc-specific-properties")
+    public static final String REST_DEPENDENCY_RANKS = "restDependencyRanks.";
+
+    @ConfigValue( docIndex = "dep-manip.html#rest-pnc-specific-properties")
+    public static final String REST_DEPENDENCY_ALLOW_LIST = "restDependencyAllowList.";
```
rnc, line 86: Why is there global as well as this property but both use the same value?

PR #994, opened 2023-08-22
Clarify docs for destination of dependencyInjection GAVs

ASzc:
The way the documentation is worded, it makes it sound like `-DdependencyInjection` adds a real dependency entry to the POMs. In fact it adds a dependencyManagement entry, as seen in the code here:

https://github.com/release-engineering/pom-manipulation-ext/blob/main/core/src/main/java/org/commonjava/maven/ext/core/impl/DependencyInjectionManipulator.java#L95


rnc:
Thanks!

PR #1130, opened 2026-04-10
NCLSUP1410 Handle properties in GAV 

rnc:
This refactors the Project.java to always have a MavenSession ; this simplifies many of the callers as they don't need to pass in a session (e.g. `currentProject.getResolved*`) and the session can be used then to return a fully resolved artifactId:groupId:version. This is critical when writing the json report or collecting versions to send to the REST API.


akridl (approved):
Thank you @rnc for the (refactor and) fix!!

PR #1136, opened 2026-04-23
NCL-9664 Move to using jboss-parent

rnc:
This moves from using redhat-releng-tools parent pom to jboss-parent.

The only issue is that I have to manually downgrade the plugins in [`69daa09` (this PR)](https://github.com/project-ncl/pom-manipulation-ext/pull/1136/commits/69daa09cd82bfc57598c3e97c9e6648d114d24a8) as many require >= Maven 3.6.3. Its not possible to activate a profile based upon the version of Maven. In [`79fbe32` (this PR)](https://github.com/project-ncl/pom-manipulation-ext/pull/1136/commits/79fbe32c397e05dcd70163b1e987a23750609ba9) I've tweaked the CI so it uses a profile to downgrade the plugins for early maven versions. This will also work locally (`-Pearly-maven`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (C) 2012 Red Hat, Inc.
+    Copyright © 2012 Red Hat, Inc.
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
     You may obtain a copy of the License at
 
-            http://www.apache.org/licenses/LICENSE-2.0
+        http://www.apache.org/licenses/LICENSE-2.0
 
     Unless required by applicable law or agreed to in writing, software
     distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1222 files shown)


pom.xml
```diff
@@ -81,6 +81,29 @@
...
     <maven.compiler.source>${javaVersion}</maven.compiler.source>
     <maven.compiler.target>${javaVersion}</maven.compiler.target>
+    <maven.compiler.testTarget>${javaVersion}</maven.compiler.testTarget>
+    <maven.compiler.testSource>${javaVersion}</maven.compiler.testSource>
+    <maven.compiler.release>8</maven.compiler.release>
+    <maven.compiler.testRelease>8</maven.compiler.testRelease>
+
+    <!-- Override enforcer requirements -->
+    <maven.min.version>3.3.9</maven.min.version>
+    <jdk.min.version>11</jdk.min.version>
```
akridl, line 91: how does these 2 work? is it `maven-enforcer-plugin` which reads these properties?

akridl (approved):
Thank you for moving the parent! (and also setting up releasing)

pom.xml
rnc: Correct, in https://github.com/jboss/jboss-parent-pom/blob/main/pom.xml#L782 and https://github.com/jboss/jboss-parent-pom/blob/main/pom.xml#L768 respectively.

PR #1186, opened 2026-08-24
NCLSUP-1478 Add a suppressManifestComment option

cunningt:
NCLSUP-1478 Add a suppressManifestComment option that will suppress the printing of the footer manifest comment "Modified by POM Manipulation Extension for Maven".

Example 

```
</project>
<!--
Modified by POM Manipulation Extension for Maven 5.6-SNAPSHOT ( SHA: e8a5fe82095d0ca622ab4efd37afd1b8d9ac95c7 )
-->
```

### DIFF
--- a/io/src/main/java/org/jboss/pnc/mavenmanipulator/io/PomIO.java
+++ b/io/src/main/java/org/jboss/pnc/mavenmanipulator/io/PomIO.java
@@ -48,6 +48,7 @@ import org.apache.maven.shared.release.transform.jdom2.JDomModelETLFactory;
 import org.codehaus.plexus.util.xml.pull.XmlPullParserException;
 import org.commonjava.atlas.maven.ident.ref.ProjectVersionRef;
 import org.commonjava.maven.galley.maven.parse.PomPeek;
+import org.jboss.pnc.mavenmanipulator.annotation.ConfigValue;
 import org.jboss.pnc.mavenmanipulator.common.exception.ManipulationException;
 import org.jboss.pnc.mavenmanipulator.common.jdom.JDOMModelConverter;
 import org.jboss.pnc.mavenmanipulator.common.model.Project;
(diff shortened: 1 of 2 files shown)


cunningt:
@rnc

io/src/main/java/org/jboss/pnc/mavenmanipulator/io/PomIO.java
```diff
@@ -78,18 +91,22 @@ public class PomIO {
...
     private final boolean parsePomTemplates;
 
     private String manifestComment;
 
     @Inject
     public PomIO(MavenSessionHandler handler) {
+        this.handler = handler;
         parsePomTemplates = Boolean.parseBoolean(
                 handler.getUserProperties().getProperty(PARSE_POM_TEMPLATES, "true"));
     }
```
rnc, line 105: Instead of storing the handler, why not just store a boolean like the `parsePomTemplates` ?

cunningt: @rnc It doesn't seem to work for me with just the boolean and I am not sure that parsePomTemplates works - I'm trying to verify locally 

`java -jar /Users/tcunning/src/community/pom-manipulation-ext/cli/target/pom-manipulation-cli-5.6-SNAPSHOT.jar -f /tmp/pme-test/pom.xml -DversionSuffix=redhat-1 -DsuppressManifestComment=true` 

Without the maven handler, I'm still seeing manifest comments.    

Here's what Claude is telling me about that

> PomIO is looked up (and constructed) at line 355, before the MavenSession with user properties is set on ManipulationSession at line 402. At construction time, ManipulationSession.mavenSession is null, so getUserProperties() returns an empty Properties — and suppressManifestComment defaults to false.

> This is why the original code stored the handler reference and read the property lazily in write() — by that point the session is fully initialized. The parsePomTemplates property has the same bug but nobody notices because its default (true) is the normal desired behavior.

rnc: @cunningt Huh, good spot! Bob agrees with you - the creation in the Cli code versus the instantiation of the session later on. And the same applies when its running as an extension for that matter. 

`parsePomTemplates` was added a long time ago and at some point I meant to remove it. Its an undocumented flag that was intended to be removed 4 years ago - feel free to remove it now if it would simplify the code.

rnc:
@cunningt Can you also add a PR to the gh-pages branch to document this please?

cunningt:
> @cunningt Can you also add a PR to the gh-pages branch to document this please?

https://github.com/project-ncl/pom-manipulation-ext/pull/1188

rnc (approved):

PR #1201, opened 2026-08-27
NCLSUP-1482 Write json file if no changes

rnc:
This will need a doc update. By default there a no user-facing changes. If `reportJsonUnchanged` is true, then the report json file will be written regardless of whether there are changes or not.


akridl (approved):
Pasting part of my comment from the jira issue:
> Setting reportJsonUnchanged to true makes a little more sense to me. But even in case it will be false by default, it’s fine: I’ll just provide -DreportJsonUnchanged=true as a default value from reqour.

otherwise, everything looks fine to me! thanks for the addition of this feature!!

PR #1206, opened 2026-09-04
fix: NPE in PluginManipulator when plugin declares no groupId or version

vibe13:
## Problem

When a `<plugin>` element is declared in `<pluginManagement>` (or `<plugins>`) with **only** `<artifactId>` and no `<groupId>` or `<version>`, PME throws a `NullPointerException` and the alignment fails.

Example POM fragment that triggers the issue:

```xml
<plugin>
  <artifactId>maven-jar-plugin</artifactId>
  <configuration>
    <archive>
      <manifestEntries>
        <Implementation-Vendor-Id>org.apache</Implementation-Vendor-Id>
      </manifestEntries>
    </archive>
  </configuration>
</plugin>
```

Stack trace:
```
java.lang.NullPointerException
    at org.jboss.pnc.mavenmanipulator.core.impl.PluginManipulator.applyOverrides(PluginManipulator.java:381)
    at org.jboss.pnc.mavenmanipulator.core.impl.PluginManipulator.apply(PluginManipulator.java:294)
    at org.jboss.pnc.mavenmanipulator.core.impl.PluginManipulator.applyChanges(PluginManipulator.java:135)
```

## Root Cause

Maven allows omitting `<groupId>` (defaults to `org.apache.maven.plugins`) and `<version>` in plugin declarations.

In `PluginManipulator.applyOverrides()`, the `pluginsByGA` map is built from resolved `ProjectVersionRef` keys. When a well-known plugin such as `maven-jar-plugin` is listed without a version, `resolvePlugins()` fills in the default version (`2.4`) from `StandardMaven350PluginDefaults` for the **key** — but the **value** stored in the map is the raw Maven `Plugin` object, which still carries `version=null` from the original model.

When a REST override is found for that plugin, line 381 called:

```java
if (plugin.getVersion().equals(Version.PROJECT_VERSION)  // NPE: plugin.getVersion() is null
```

## Fix

Guard both `plugin.getVersion()` usages on line 381–382 with a `!= null` check, consistent with how `oldVersion` is already guarded at lines 512 and 518:

```java
if (plugin.getVersion() != null && plugin.getVersion().equals(Version.PROJECT_VERSION)
        || (plugin.getVersion() != null && plugin.getVersion().contains("$")
                && project.getVersion().equals(oldValue))) {
```

When `plugin.getVersion()` is `null`, neither the `${project.version}` skip nor the strict-mode check applies, so the code falls through to the version-setting block at line 521 (`plugin.setVersion(newValue)`), which is the correct behaviour.

## Test

Added `PluginManipulatorTest.pluginWithoutGroupIdAndVersionDoesNotThrowNPE()` which:
1. Creates a project with a `<pluginManagement>` block containing `<maven-jar-plugin>` with no `groupId` and no `version`
2. Injects a REST override for `org.apache.maven.plugins:maven-jar-plugin`
3. Verifies no NPE is thrown and the version override is applied correctly

## Observed in production

Reproducer: `org.apache.httpcomponents:httpcomponents-parent:14` (tag `rel/v14`) aligned with PME 5.7 in LIGHTWELL/RHLW mode.

### DIFF
--- a/core/src/test/java/org/jboss/pnc/mavenmanipulator/core/impl/PluginManipulatorTest.java
+++ b/core/src/test/java/org/jboss/pnc/mavenmanipulator/core/impl/PluginManipulatorTest.java
@@ -17,6 +17,7 @@ package org.jboss.pnc.mavenmanipulator.core.impl;
 
 import static org.jboss.pnc.mavenmanipulator.core.fixture.TestUtils.createSession;
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
 
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/jboss/pnc/mavenmanipulator/core/impl/PluginManipulatorTest.java
```diff
@@ -0,0 +1,110 @@
...
+     * @see <a href="https://github.com/release-engineering/pom-manipulation-ext/issues">PME issue</a>
+     */
+    @Test
+    public void pluginWithoutGroupIdAndVersionDoesNotThrowNPE() throws ManipulationException {
+        // Arrange: session using REST as plugin source so the REST override is active.
+        // Disable strict alignment to match the production configuration in which the NPE was
+        // observed (strict alignment is not in the pncDefaultAlignmentParameters for LIGHTWELL builds).
+        Properties props = new Properties();
+        props.setProperty("pluginSource", "REST");
+        props.setProperty("strictAlignment", "false");
```
rnc, line 63: Could we keep this as true - that is the default we use

vibe13: added a new test below

rnc (approved):

PR #1207, opened 2026-09-04
NCL-9969 Add enforceVersionPrefix

rnc:


core/src/main/java/org/jboss/pnc/mavenmanipulator/core/impl/VersionCalculator.java
```diff
@@ -324,4 +327,36 @@ protected static String handleAlternate(VersioningState state, String version) {
...
+     */
+    public static String enforceVersionPrefix(VersioningState state, String version) {
+        final String reference = state.getEnforceVersionPrefix();
+        if (isEmpty(reference)) {
+            return version;
+        }
+        // Match the reference preceded by '.' or '-' and followed by '-' and one or more digits,
+        // anywhere in the version string — the match need not be terminal.
+        final Pattern alreadyPresent = Pattern.compile(
+                "(?:[.-])" + Pattern.quote(reference) + "-[0-9]+");
```
vibe13, line 354: Looks good! Why are you using `Pattern.quote(reference)` here but `final String suffixStripRegExp = "(.*)([.|-])(" + suffix + "-\\d+)";` in the `handleAlternate` method? Should they be the same or are they handling different cases?

vibe13 (approved):
Just one small question. Looks good!! Thanks! :-)

core/src/main/java/org/jboss/pnc/mavenmanipulator/core/impl/VersionCalculator.java
rnc: They handle different scenarios. [Alternate suffix](https://project-ncl.github.io/pom-manipulation-ext/guide/project-version-manip.html#alternate-suffix-handling) is to allow to handle alignment when DA returns both e.g. temporary-rh and rh sufffixes (See NCL-4165). In contrast the enforce handling is to ensure a certain prefix is there.

vibe13: ah ok thanks! +1 :-)

